PR #368, opened 2014-11-15
Non-ajax Select2 component

lehvolk:
Now select2 component can be non-ajax.


tgoetz:
IMHO this implementation breaks current implementations, see current discussion here: http://markmail.org/thread/u2ccale6aubcmm5p

PR #377, opened 2015-01-11
More OpenLayers3 Changes

cmiles74:
This pull request includes some significant changes, the biggest is that all of the Javascript variables are now stored in one global array; the map can now be added to a panel via AJAX (earlier releases required the map to be present when the page was constructed). Performance of the Modify interaction has been improved, changes accumulate and Wicket is notified once the feature is no longer in the process of being modified.

I am not sure how to exclude the couple of commits where I altered the Travis configuration. Sorry about that!


martin-g:
Thanks!

klopfdreh:
Is it right that the changes of https://github.com/cmiles74/core/commit/886d87a5fdba2a7bbaa01e62e9991f0f42ea9b07 and https://github.com/cmiles74/core/commit/086f47747ad3f91a61008dad045aba9c6ba54d94 went into the jdk-1.6-parent - seems to be a mistake isn't it?

martin-g:
Yes, it is !
In master branch we should use jdk-1.7-parent.

PR #386, opened 2015-03-13
Altered to Use Wicket Webjars for OpenLayers Resources

cmiles74:
With this change the library will now use Wicket Webjars to pull in the OpenLayers 3 resources. This should resolve the issue where people set their application to HTTPS and the library continues to fetch resources via HTTP.

I also removed a call to "dispose" on interactions that causes an error when used against OpenLayers 3.3.x.

### DIFF
--- a/jdk-1.7-parent/openlayers3-parent/openlayers3-examples/src/main/java/org/wicketstuff/openlayers3/examples/WicketApplication.java
+++ b/jdk-1.7-parent/openlayers3-parent/openlayers3-examples/src/main/java/org/wicketstuff/openlayers3/examples/WicketApplication.java
@@ -6,6 +6,8 @@ import org.apache.wicket.Page;
 import org.apache.wicket.RuntimeConfigurationType;
 import org.apache.wicket.protocol.http.WebApplication;
 import org.wicketstuff.annotation.scan.AnnotatedMountScanner;
+import de.agilecoders.wicket.webjars.WicketWebjars;
+import de.agilecoders.wicket.webjars.settings.WebjarsSettings;
 
 /**
  * Provides an application that demonstrates the OpenLayers3 map.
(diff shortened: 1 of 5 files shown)


jdk-1.7-parent/openlayers3-parent/openlayers3-examples/src/main/java/org/wicketstuff/openlayers3/examples/WicketApplication.java
```diff
@@ -26,9 +28,13 @@ protected void init() {
         // scan for annotations
         new AnnotatedMountScanner().scanPackage("org.wicketstuff.openlayers3.examples").mount(this);
 
+		// setup webjars
+		WebjarsSettings webjarsSettings = new WebjarsSettings();
```
martin-g: whitespace issue

jdk-1.7-parent/openlayers3-parent/openlayers3/src/main/java/org/wicketstuff/openlayers3/DefaultOpenLayersMap.java
```diff
@@ -27,12 +28,12 @@ public DefaultOpenLayersMap(final String id, final IModel<Map> model) {
 
     @Override
     public void renderHead(final IHeaderResponse response) {
-        response.render(CssHeaderItem.forUrl("http://ol3js.org/en/master/css/ol.css"));
```
martin-g: You can also use `CssHeaderItem.forUrl("//ol3js.org/en/master/css/ol.css")` to make it scheme relative.

martin-g:
Two minor notes. Everything else looks good!
Let me know when you are ready and I'll merge it!
Thanks @cmiles74 !

jdk-1.7-parent/openlayers3-parent/openlayers3/src/main/java/org/wicketstuff/openlayers3/DefaultOpenLayersMap.java
cmiles74: Now that it's using webjars instead, I think we're good no matter what. :-)

cmiles74:
I fixed that whitespace issue as well, I think this is all set.

PR #387, opened 2015-03-13
Wicket 6.x: Altered to Use Wicket Webjars for OpenLayers Resources

cmiles74:
These are the same changes as the pull request against the master branch.

With this change the library will now use Wicket Webjars to pull in the OpenLayers 3 resources. This should resolve the issue where people set their application to HTTPS and the library continues to fetch resources via HTTP.

I also removed a call to "dispose" on interactions that causes an error when used against OpenLayers 3.3.x.


martin-g:
Thanks!

PR #391, opened 2015-04-13
Reformat select2 code

shuraa:
... according to http://mvnrepository.com/artifact/org.apache.wicket/wicket-eclipse-settings


reiern70:
Wouldn't this make tracking issues a lot more difficult?

solomax:
I personally will better have well formatted code, not sure if I need to start discussion on that ....
should I?

reiern70:
let's see what @martin-g  and other commiters thinks ;-) 

I still have some "private" version of select2 with things that might be good additions. I will wait for this to be applied or rejected to create a PR adding those.

martin-g:
Formatting only changes are actually what I prefer.
I do not like PRs with fix/improvement that also have formatting noise.

solomax:
Great!
Thanks a lot Martin :)

PR #392, opened 2015-04-13
Reformat select2 code

shuraa:
... according to http://mvnrepository.com/artifact/org.apache.wicket/wicket-eclipse-settings


martin-g:
Thanks!

Issue #390, opened 2015-04-09
Select2Choice lose display data after ajax update


egg1501:
this issue was already fixed (https://groups.google.com/forum/#!topic/select2/9nFPPjFs4-w) in wicket-select2 (https://github.com/ivaynberg/wicket-select2).

With wicket-select2 (https://github.com/ivaynberg/wicket-select2). everything works fine.

Only problems with wicketstuff-select2!

solomax:
Will try to address it ASAP
Thanks for the report

solomax:
could you please describe what is wrong more detailed?
the links you have provided are for select2 v1, we are currently using: v3.5.2

simple quickstart might help to reproduce/fix the issue
Thanks in advance!

egg1501:
Quickstart https://github.com/egg1501/select2-ajax-update-issue

If i try to update the components in onError method of the button

@Override
			protected void onError(final AjaxRequestTarget target, Form<?> form) {
form.visitFormComponents(new IVisitor<FormComponent<?>, Object>() {
					@Override
					public void component(final FormComponent<?> component, final IVisit<Object> objectIVisit) {		target.add(component);
						}
				});)
}

the select2 component lose the display value (getConvertedInput() is not null!)

egg1501:
Did you found something?

shuraa:
Hi ! You can clone my repo https://github.com/shuraa/core/tree/wicket-6.x . I've reworked Select2Choice and Select2MultiChoice in parts there value is obtained and converted into user type. I hope, there is no more issues with Ajax updates :) After while I'll create PR to wicketstuff/core.

solomax:
Hello @egg1501, unfortunately I had no time to take a look at it?

@shuraa Thanks for investigating it! can you please split PR into 2 parts:
1) issue fix
2) formatting fix, BTW are you using http://mvnrepository.com/artifact/org.apache.wicket/wicket-eclipse-settings while applying formatting?

shuraa:
@solomax I've used formatting rules from here https://github.com/shuraa/core/tree/master/config/eclipse . Should I use another rules?

solomax:
Wicket team released eclipse settings as separate module, I would use this version.
Here is the link to the VOTE: http://apache-wicket.1842946.n4.nabble.com/VOTE-Release-wicket-eclipse-settings-1-td4669969.html

shuraa:
Okay. I'll drop commits with formatting and create PR.

PS
[...]

PR #394, opened 2015-04-16
Select2Choice and Select2MultiChoice have to retain value...

shuraa:
Issue #390 should be fixed. I've added several tests that covers several scenarios:
* select2 retain value after page refresh
* select2 retain value if form wasn't validated


jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -137,19 +146,21 @@ public AbstractSelect2Choice(String id, IModel<M> model, List<T> collection)
 	 *            renderer list item
 	 * @see HiddenField#HiddenField(String, IModel)
 	 */
+	// will be dropped in 7.0
+	@Deprecated
 	public AbstractSelect2Choice(String id, IModel<M> model, List<T> choices,
-		ChoiceRenderer<T> renderer)
+	                             ChoiceRenderer<T> renderer)
```
solomax: not sure why change like this is necessary, seems to be redundant
can you remove it? (lots in this file)

solomax:
@martin-g you usually add "whitespace police" comments :) is it OK to commit PR with such witespace changes or it need to be additionally cleaned up?

shuraa:
More cleanups in this PR https://github.com/wicketstuff/core/pull/395

PR #395, opened 2015-04-19
Fix issue 390 (rework Select2Choice and Select2MultiChoice)

shuraa:
- Added Select2ChoiceTest that reproduces issue with loosing value by Se...lect2Choice after form submit.
- Override convertInput() and convertValue() in order to convert and process properly Select2Choice validation phase.
- Override getInputAsArray() for Select2MultiChoice in order to satisfy FormComponent's contract. Implement convertValue() with ChoiceProvider.

PS
There are still some formatting changes. I have no idea that I've done wrong :(

### DIFF
--- a/jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
+++ b/jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
@@ -55,7 +55,7 @@ abstract class AbstractSelect2Choice<T, M> extends HiddenField<M> implements IRe
 	private List<T> choices;
 	private ChoiceRenderer<T> renderer;
 
-	private transient boolean convertValuePerformed = false;
+	private boolean convertInputPerformed = false;
 
 	/**
 	 * Constructor
@@ -220,7 +220,7 @@ abstract class AbstractSelect2Choice<T, M> extends HiddenField<M> implements IRe
 		// AbstractSelect2Choice uses ChoiceProvider to convert IDS into objects.
 		// The #getConverter() method is not supported by Select2Choice.
 		setConvertedInput(convertValue(getInputAsArray()));
-		convertValuePerformed = true;
+		convertInputPerformed = true;
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -54,9 +55,11 @@
 	private List<T> choices;
 	private ChoiceRenderer<T> renderer;
 
+	private transient boolean convertValuePerformed = false;
```
martin-g: No need to be `transient`. The first is reset to `false` in `#onDetach()`.
Additionally: the page is stored in the HttpSession and the next request may read it as a live instance from there, so `transient` won't help. It will help only if the page is deeper in the history stack, i.e. in the disk store, serialized as byte[].

martin-g: The name of the field probably should be `convertInputPerformed` to be in sync with the name of the method.

jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2MultiChoice.java
```diff
@@ -193,10 +168,8 @@ protected void renderInitializationScript(IHeaderResponse response)
 					}
 				}
 			}
-
-			return result;
+			return result.toArray(new String[result.size()]);
 		}
-
-		return Arrays.asList(input.split(","));
+		return input.split(","); // select2 uses comma as value separator
```
martin-g: Use wicket-util's Strings#split() for better performance

jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
shuraa: Fixed.

jdk-1.6-parent/select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2MultiChoice.java
shuraa: Done. Thanks!

solomax:
Thanks for the patch!

PR #400, opened 2015-04-22
Wicket 6.x: Added Extent object, Better Click Handling, Cluster Demo Warning

cmiles74:
I coded up an Extent object that may be used to specify an area (a minimum coordinate and a maximum). This extent can then be used to instruct the map on what needs to be in the viewable area; the map will make a best effort to zoom to a level that includes the full extent.

I altered the map to correctly handle single clicks (double clicks are handled by OpenLayers).

I added a warning to the cluster demo page that indicates the issue with the cluster data source and panning of the map.


martin-g:
Thank you!

PR #402, opened 2015-05-04
Revert non-ajax support

tgoetz:
As already mentioned on wicket-users mailing list (http://markmail.org/message/w2ynmeguy6szk3uf), I think the current implementation of wicket-select2 in not usable because of commit 1f7aaeb (non-ajax support):

First of all, I think there is no need for a non-ajax wicket-select2 component: if you need a non-ajax select2, simply use a DropDownChoice (or hidden textfield for multichoice) and init select2 directly using Javascript only (select2.js). I could provide a Behavior that handles the JS initialization part.

The current non-ajax implementation uses an ArrayList in addition to the already existing ChoiceProvider, but this breaks the idea behind the ChoiceProvider, now there are two separate ways to provide choices, which is very confusing. Also, users are forced to provide the (static) list of (non-ajax) choices (else the constructor throws exceptions), although they might not be interested in a non-ajax behavior.

If the PR is accepted I will also merge it to the wicket-6.x branch and see what else needs to be merged.


reiern70:
Hi,

1- This non AJAX can be achieved with a behavior attached to a drop down choice as mentioned by @tgoez

package com.vaynberg.wicket.select2;

import org.apache.wicket.Component;
import org.apache.wicket.behavior.Behavior;
import org.apache.wicket.markup.head.IHeaderResponse;
import org.apache.wicket.markup.head.OnDomReadyHeaderItem;

/**
 * Behavior that adds a select2 to a dropdown choice.
 */
public class Select2Behavior extends Behavior {

    @Override
    public void renderHead(Component component, IHeaderResponse response) {
        new Select2ResourcesBehavior().renderHead(component, response);
        StringBuilder builder = new StringBuilder("$('#").append(component.getMarkupId()).append("').select2({});");
        response.render(OnDomReadyHeaderItem.forScript(builder));
    }

    @Override
    public void bind(Component component) {
        component.setOutputMarkupId(true);
    }
}

2- What I find useful is reading content from another sources like mounted resources 

https://github.com/ivaynberg/wicket-select2/pull/45

I'm +1 to revert.

PR #414, opened 2015-07-13
Improvements removing spring core add Reflections keeping classpath compatibility.

felipecalmeida:
Hi all,

I looked forward about a JIRA to create a ticket for this component but I didn't found any. So I'm creating this PR to submit, if you guys like this modification, removing the Spring Core lib and adding Reflections lib from https://github.com/ronmamo/reflections.
The library size is less than Spring Core and keep the compatibility with the classpath when searching for annotations, classes, packages...
I did not made a performance test but i think it's pretty similar to what Spring core does.
Another benefit is that the code is less than before, easier to understand and less test class.
I deployed the openlayers3 examples https://github.com/wicketstuff/core/tree/master/jdk-1.7-parent/openlayers3-parent/openlayers3-examples to verify if this modification keep the applications working as expected.

I hope you like it.
Kind Regards,
Felipe.


jdk-1.7-parent/annotation/pom.xml
```diff
@@ -42,6 +37,11 @@
 			<groupId>log4j</groupId>
 			<artifactId>log4j</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.reflections</groupId>
+			<artifactId>reflections</artifactId>
+			<version>0.9.10</version>
```
martin-g: Please extract `<reflections.version>` in the project parent pom so it is easier to keep it up-to-date.

jdk-1.7-parent/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java
```diff
@@ -97,46 +98,16 @@
 {
 
 	/**
-	 * Get the Spring search pattern given a package name or part of a package name
-	 * 
-	 * @param packageName
-	 *            a package name
-	 * @return a Spring search pattern for the given package
-	 */
-	public String getPatternForPackage(String packageName)
```
martin-g: Maybe it would be better to mark it as deprecated and leave the old impl. Just to make migration to 7.x easier. We will drop all deprecated methods with 8.c

jdk-1.7-parent/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java
```diff
@@ -97,46 +98,16 @@
...
-	}
-
-
-	/**
-	 * Scan given a package name or part of a package name and return list of classes with MountPath
-	 * annotation.
-	 * 
-	 * @return A List of classes annotated with &#64;MountPath
-	 */
-	public List<Class<?>> getPackageMatches(String pattern)
```
martin-g: Same here.
I think it'd be better to keep these public methods implemented with Reflections.

jdk-1.7-parent/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java
```diff
@@ -97,46 +98,16 @@
...
-	{
-		return getPatternMatches(getPatternForPackage(pattern));
-	}
-
-	/**
-	 * Scan given a Spring search pattern and return list of classes with MountPath annotation.
+	 * Scan a list of classes which are annotated with MountPath
 	 * 
-	 * @return A List of classes annotated with &#64;MountPath
+	 * @param mounts
```
martin-g: The parameter name is actually `patterns`.
Please give an example what a `pattern` could look like, so it is easier for beginners what to provide.

jdk-1.7-parent/annotation/src/main/java/org/wicketstuff/annotation/scan/AnnotatedMountScanner.java
```diff
@@ -145,42 +116,7 @@ public String getPatternForPackage(String packageName)
...
-	}
-
-	/**
-	 * Scan given package name or part of a package name
-	 * 
-	 * @param packageName
-	 *            a package to scan (e.g., "org.mycompany.pages)
-	 * @return An {@link AnnotatedMountList}
-	 */
-	public AnnotatedMountList scanPackage(String packageName)
```
martin-g: These methods should be preserved. Someone may make use of them.

felipecalmeida:
Martin,

I'm going to do the modifications.
Thank's!

felipecalmeida:
I don't know if it was too much @Deprecated, just tell me where I can fix.

martin-g:
Hi Felipe,

Sorry if I wasn't clear the first time! I know that time is a precious resource for everyone and I appreciate your help!

What I meant is to try to keep as much public methods as possible, but reimplemented with Reflections. And mark them deprecated if they don't bring much value.

But now looking at the new diff I think it may be much easier just to add a new Scanner based on Reflections next to the one based on Spring. We could extract an interface for the common methods too.
I think having a separate Maven module for the new Scanner would be too much. So I suggest to make both dependencies (spring and reflections) optional. This way the application may decide which dependency to include and respectively which Scanner to use.

I understand that the current impl is not suitable for JEE applications (and even for apps using Guice!) but we have to make it as transparent as possible for all current Spring users of this library.

What do you think?

felipecalmeida:
Hi Martin,

No problem, I use wicket around a year or more and some ideas come in mind, so it's time to do some effort to help the community and It's not a problem to use me as a resource, I'm learning how to contribute and it's kind of a way to help others.

Let me see if I understand your suggestion. It could be an Interface called like IScanner that is implemented both AnnotatedMountScanner and AnnotatedReflectionMountScanner? Where AnnotatedReflectionMountScanner make use of Reflections that is optional on pom.xml (so Spring also)? But AnnotatedReflectionMountScanner can have an method scanPackage that has as parameters (String ... packages) because Reflections can handle that.

Reflections can be used with Spring, it's not a problem, just the use of classpath*: when you declare the package on the method scanPackage is not required anymore, because Reflections has ClasspathHelper inside that uses classpath (I think it's default, I was following the implementation and has something developed about that).

Am I going in the right direction?
It's not a problem if it'll not be accepted or it takes work effort, if I can help on another thing, just tell me and I'm ready for it.
Thank's!

martin-g:
Yep, everything is correct!

s/IScanner/IMountScanner/

You may add a protected factory method for the instantiation of `Reflections` instance in AnnotatedReflectionMountScanner, so application developers can use custom builder if needed.

felipecalmeida:
Martin,

Sorry, but you can cancel this PR. I made some research (and added a test that shows the issue) that Reflections can't handle with Wildcards on package name, so it is not the best choice to mount the pages. As I can see, I missed an important test before the first commit and PR. 

Anyway, I'm really sorry and thank you again for your help and time!
Kind Regards,

Felipe.

martin-g:
What exactly is not supported by Reflections?
I don't use this library in my apps but usually people use it by applying `@MountPath("/some/path")` on their pages.
The library uses Wicket's MountedMapper, not PackageMapper, so everything should be OK, IMO.

felipecalmeida:
Martin,

I found that Reflections can't handle with Wildcard, like org.*.annotation.packages. It just doesn't accept * (asterisk) on an package for scan the classes. The wicketstuff-annotation has examples with this kind of suggestion on an application and Spring can do it, but Reflections can't.
That's why Reflections really doesn't worthwhile.

martin-g:
I think this should be possible with Reflections too.
1. new Reflections("org") - so it scans everything in package `org`
2. Add a custom Scanner impl that rejects anything that doesn't have `annotation.packages` in its package name.

felipecalmeida:
Martin,

Sure, but my initial idea was that less code doing the same thing as Spring does, if I have to implement what Spring already do, I think for maintenance will be too much effort.
I thought that Reflections could do everything that Spring (talking about package scan/class scan/classpath scan/etc) does without lot of code.

martin-g:
OK. Then I'll close this issue.

felipecalmeida:
Thank you Martin!

PR #416, opened 2015-07-13
reflect type from modelobject

duesenklipper:
Building on my other pull request (#415), this adds a separate from method named fromObject(IModel<T>) that uses target.getObject().getClass() to determine the target type. This is useful for the (pretty common) case of models with an erased type parameter. I'm using a separate method rather than doing this in from(IModel<T>) to not change existing behavior and to avoid calling methods on the target when the user does not expect it.

### DIFF
--- a/jdk-1.7-parent/wicketstuff-lazymodel/src/main/java/org/wicketstuff/lazymodel/LazyModel.java
+++ b/jdk-1.7-parent/wicketstuff-lazymodel/src/main/java/org/wicketstuff/lazymodel/LazyModel.java
@@ -556,6 +556,19 @@ public class LazyModel<T> implements IModel<T>, IObjectClassAwareModel<T>,
 	}
 
 	/**
+	 * Shortcut for {@link #from(Class)} and {@link #bind(Object)}. Use this for quick model building on top of a
+	 * type-erased model.
+	 *
+	 * @param target The model to bind to.
+	 * @param type The type parameter for that model.
+	 * @param <T>
+	 * @return a result proxy for further evaluation.
+	 */
+	public static <T> T from(IModel<T> target, Class<T> type) {
+		return (T) new BoundEvaluation<T>(type, target).proxy();
+	}
+
+	/**
 	 * Start a lazy evaluation.
 	 * 
 	 * @param target
--- a/jdk-1.7-parent/wicketstuff-lazymodel/src/main/java/org/wicketstuff/lazymodel/LazyModel.java
+++ b/jdk-1.7-parent/wicketstuff-lazymodel/src/main/java/org/wicketstuff/lazymodel/LazyModel.java
@@ -563,9 +563,23 @@ public class LazyModel<T> implements IModel<T>, IObjectClassAwareModel<T>,
 	 * @param type The type parameter for that model.
 	 * @param <T>
 	 * @return a result proxy for further evaluation.
+	 * @see #fromObject(IModel)
 	 */
 	public static <T> T from(IModel<T> target, Class<T> type) {
 		return (T) new BoundEvaluation<T>(type, target).proxy();
+	}
+
+	/**
+	 * Start a lazy evaluation, using the reflected runtime type of the given model's object. Use this for terse model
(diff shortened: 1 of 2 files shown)


jdk-1.7-parent/wicketstuff-lazymodel/src/main/java/org/wicketstuff/lazymodel/LazyModel.java
```diff
@@ -556,6 +556,33 @@ private void fillArguments(Object[] args) {
...
+	 * Start a lazy evaluation, using the reflected runtime type of the given model's object. Use this for terse model
+	 * building on top of a type erased model.
+	 * @param target The model to bind to.
+	 * @param <T>
+	 * @return a result proxy for further evaluation.
+	 * @see #from(IModel, Class)
+	 */
+	public static <T> T fromObject(IModel<T> target)
+	{
+		return (T) new BoundEvaluation<T>(target.getObject().getClass(), target).proxy();
```
martin-g: This eagerly evaluates the model object. Maybe it is not what you really want.

svenmeier: Apparently this is exactly what Carl-Eric wants to do here.

As convenient as this sounds, I don't think we should have such a method in the API: it's too easy to dereference the model inadvertently.

duesenklipper: Yes, that is indeed exactly what I want to do. 

I agree that it should not be the default behavior: That is why I put it in a separate, non-overloaded method, so you'd have to explicitly and knowingly call it. The only thing with which I'm not entirely happy is the name "fromObject". I'm open to better ideas.

duesenklipper: Maybe "eagerlyFrom"?

svenmeier: Yes "eagerlyFrom" would definitely go into the right direction:

I find it very important that LazyModel never accesses the model object during evaluation - I'm afraid adding such a possibility will lead to people calling the wrong method.

Which model implementations are you using in your project, so you have to provide the object class by yourself?

svenmeier: See WICKET-5950 too.

duesenklipper: I do understand the concern. The model object shouldn't be accessed *unless you know what you're doing and why it's bad*. 

The problem is that I'm dealing with a legacy codebase that has stuff like CompoundPropertyModels and plain Models all over the place. I'd like to gradually introduce LazyModel without having to change everything.

Hm. Maybe provide something like eagerlyFrom in a separate class so it would have to be a separate import? Something like "LegacySupport.eagerlyFrom" or so?

martin-g:
```java
import org.wicketstuff.lazymodel.BadPractice

...
... = BadPractice.fromModelObject(anotherModel);
...
```

:-)

duesenklipper: I wouldn't go *that* far ;-)

I think together with WICKET-5950 calling it LegacySupport or ErasureModelSupport or something along those lines would be good enough.

svenmeier:
If we add the method to a separate class, then I really don't see the need to have any support for this usecase in wicketstuff. You can easily roll your own legacy method.

duesenklipper:
Fair enough.

PR #418, opened 2015-08-13
Fixed attribute in LoginButton constructor to get facebook permissions to work correctly.

Staretta:
I just started using the wicket facebook stuff, and found that it works when requesting default permissions, but when I add additional permissions they never seem to get added. After a bit of reading, I saw a comment that said they changed it from data-perms to data-scope. So I changed that and it seems to work for me now.


shapewerk:
Having the same problem - still.  Only when replacing perms with scope attribute the additional permission is added, otherwise it's not working. Please fix.

martin-g:
Please send a Pull Request!
Thank you!

PR #423, opened 2015-10-05
selectize.js Integration

klopfdreh:
See examples:

https://github.com/klopfdreh/core/blob/435e7aa23f6e45daeb7b0b97ad7da7a2bd8571b2/wicketstuff-selectize-parent/wicketstuff-selectize-examples/src/main/java/org/wicketstuff/selectize/TestPage.java


klopfdreh:
If everything is ok - we have to add the module as submodule.

wicketstuff-selectize-parent/wicketstuff-selectize-examples/src/main/java/org/wicketstuff/selectize/TestPage.java
```diff
@@ -0,0 +1,77 @@
...
+
+		// Select with ajax
+		Selectize selectize4 = new Selectize("selectize4")
+		{
+			private static final long serialVersionUID = 1L;
+
+			@Override
+			protected SelectizeResponse response(String search)
+			{
+				if (search != null && search.equals("test"))
```
martin-g: Could be simplified with: `if ("test".equals(search))`

martin-g:
:+1: 
Just mention somewhere in the javadoc that the template should be Handlebars.

klopfdreh:
I am going to write some javadoc tomorrow evening and after that I think we can integrate it.

PR #426, opened 2015-10-12
Mark slf4j-log4j12 as scope=test. Fixed #425

### LINKED ISSUE #425: wicketstuff-annotation 7.0.0+ should mark dependencies slf4j-log4j12 and log4j as optional=true or scope=test, depend on slf4j-api instead
If these are used by tests only, they should be marked as scope=test.

If log4j is used in runtime code (why?), they should be marked as optional=true.

To ensure slf4j as dependency, add this as normal dependency: 

    <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-api</artifactId>
        <version>1.7.12</version>
    </dependency>

ceefour:


martin-g:
Thanks!

PR #427, opened 2015-10-12
removed unused logging dependencies

AlexHaxe:
causes a log4j dependency that messes up your slf4j logging setup


martin-g:
Thanks!

PR #436, opened 2015-11-18
Options changed with select2 4

selckin:


solomax:
Thanks!

select2-parent/select2/src/main/java/org/wicketstuff/select2/Settings.java
```diff
@@ -58,7 +57,7 @@
 
 	private AjaxSettings ajax;
 	private String data;
-	private String tags;
+	private Boolean tags;
```
martin-g: Any reason why it should be a **B**oolean and not **b**oolean?

solomax: Other boolean properties are Boolean
not sure why is it necessary

martin-g: Unless `null` is a required third value I don't see a reason to use **B**oolean.

solomax: I'll try to check documentation for default values and fix this (no estimates unfortunately :(( )

solomax: Actually all Boolean options are JS Objects
so I believe it is safe to change all Boolean to be boolean.

I'll change it right now

PR #438, opened 2015-12-03
Fixed GMAP_API_URL to use &amp; instead of &

VaclavC:
Changed GMAP_API_URL to use &amp; instead of &. Old version caused errors in W3C validator.


martin-g:
What is the generated HTML?
I need to see the HTML snippet with this url in it.
Because I'm afraid the new value will not actually work.

VaclavC:
`<script type="text/javascript" src="http://maps.google.com/maps/api/js?v=3&amp;sensor=false"></script>`

Working for me and it is OK in W3C validator

VaclavC:
You can see it here: http://test.udi-byty.cz (access is test/test)

martin-g:
OK. Thanks!
But we have to merge this to `wicket-6.x`. 6.19.0 is already released...
Sorry!
Please create a new Pull Request.

PR #440, opened 2015-12-05
Added expire for redis session

alikhon:
Session element will expire after TTL duration and removes from Redis database.


martin-g:
Thank you for your contribution!
It will be part of WicketStuff 7.2.0.

PR #441, opened 2015-12-15
Should not share userProvidedResourceReferences between instances

qlp:
JQueryAjaxBehavior has a static [userProvidedResourceReferences](https://github.com/wicketstuff/core/blob/master/jwicket-parent/jwicket-core/src/main/java/org/wicketstuff/jwicket/JQueryAjaxBehavior.java#L212), which causes problems:
- [Every time a JQueryUiTooltip is instantiated](https://github.com/wicketstuff/core/blob/master/jwicket-parent/jwicket-ui/jwicket-ui-tooltip/src/main/java/org/wicketstuff/jwicket/ui/tooltip/JQueryUiTooltip.java#L104) a JQueryResourceReference.get() instance is added (memory consumption increases)
- Adding elements to the static member is not thread safe. Instantiating JQueryUiTooltips concurrently results in a 'null' references in the list. Having one 'null' references leads to an assertion [while rendering the behavior](https://github.com/wicketstuff/core/blob/master/jwicket-parent/jwicket-core/src/main/java/org/wicketstuff/jwicket/JQueryAjaxBehavior.java#L186) (with an application restart as only 'solution').

I do not know the reason for this decision? Please have a look. This request is based on the master, but we're still on version 6. Please apply this fix also on the other branches, or let me know when I can help.

### DIFF
--- a/jwicket-parent/jwicket-core/src/main/java/org/wicketstuff/jwicket/JQueryAjaxBehavior.java
+++ b/jwicket-parent/jwicket-core/src/main/java/org/wicketstuff/jwicket/JQueryAjaxBehavior.java
@@ -209,13 +209,13 @@ public abstract class JQueryAjaxBehavior extends AbstractDefaultAjaxBehavior {
 		}
 	}
 
-    private static final List<JavaScriptResourceReference> userProvidedResourceReferences = new ArrayList<JavaScriptResourceReference>();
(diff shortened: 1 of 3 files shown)


martin-g:
Merged it to wicket-6.x too!
Thank you for the contribution!

PR #442, opened 2015-12-15
Fix usage of diamond operator in wicket 6

qlp:
The port to wicket-6.x of https://github.com/wicketstuff/core/pull/441 doesn't work. The diamond operator cannot be used in Java 6.


martin-g:
Thanks!

PR #443, opened 2015-12-29
Add support for stateless behavior

reiern70:


select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -54,7 +56,7 @@
 
 	private final Settings settings = new Settings();
 
-	private ChoiceProvider<T> provider;
+	private ChoiceProvider<T>provider;
```
solomax: this change looks strange, I believe whitespace is missing

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -273,7 +275,14 @@ protected void onInitialize()
 	protected void onConfigure()
 	{
 		super.onConfigure();
-		if (isAjax())
+        if(getSettings().isStateless())
```
solomax: seems to be mixture of space/tab delimiters
all code has spaces, so I would leave it consistent

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -309,63 +322,69 @@ public boolean isAjax()
...
-		page -= 1;
+        Response<T> response = new Response<>();
+        provider.query(term, page, response);
 
-		Response<T> response = new Response<>();
-		getProvider().query(term, page, response);
+        // jsonize and write out the choices to the response
 
-		// jsonize and write out the choices to the response
+        OutputStreamWriter out = new OutputStreamWriter(outputStream, request.getCharset());
```
solomax: seems to be resource leak: out never closes (also was in original code)

select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Choice.java
```diff
@@ -77,7 +77,13 @@ public Select2Choice(String id, ChoiceProvider<T> provider)
 		super(id, provider);
 	}
 
-	@Override
+    // needed to be able override the class on Scala
```
solomax: This method seems to be useless

reiern70: Its is needed because otherwise I cannot override method in scala. If you think it is too offending I can get rid of it and roll my own class on top of it and extend for there.

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
reiern70: Where exactly is resource closed on original code? I cannot spot it :-(

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
reiern70: I will clan up white spaces.

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
martin-g: It is good idea to close it!

select2-parent/select2/src/main/java/org/wicketstuff/select2/JsonResourceReference.java
```diff
@@ -0,0 +1,65 @@
...
+ * 
+ * @author Ernesto Reinaldo Barreiro (reiern70@mail.com).
+ */
+public abstract class JsonResourceReference<T> extends ResourceReference 
+{
+	private static final long serialVersionUID = 1L;
+
+	public JsonResourceReference() 
+    {
+		super(JsonResourceReference.class, "images");
```
martin-g: `images`?
Better pass the name as an argument.

select2-parent/select2/src/main/java/org/wicketstuff/select2/JsonResourceReference.java
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	@Override
+	public IResource getResource() 
+    {
+	    ByteArrayOutputStream webResponse = new ByteArrayOutputStream();
+	    AbstractSelect2Choice.generateJSON(getChoiceProvider(), webResponse);
+	    StringResourceStream resourceStream = new StringResourceStream(webResponse.toString(), "application/json");
```
martin-g: This uses the platform's default encoding. 
I think it is safer to use `.toString("UTF-8")`

select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Behavior.java
```diff
@@ -0,0 +1,27 @@
+package org.wicketstuff.select2;
+
+import org.apache.wicket.Component;
+import org.apache.wicket.behavior.Behavior;
+import org.apache.wicket.markup.head.IHeaderResponse;
+import org.apache.wicket.markup.head.OnDomReadyHeaderItem;
+
+/**
+ * Behavior that adds a select2 to a dropdown choice.
```
martin-g: There is no restriction to be `DropDownChoice` in the code below.

select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Choice.java
martin-g: What exactly is the issue in Scala?

reiern70: In order to override a method it should be defined by the parent class. I can go around that by rolling my own class the defines onInitialize

select2-parent/select2/src/main/java/org/wicketstuff/select2/JsonResourceReference.java
reiern70: Probably the result of some copy and paste.

select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
reiern70: ok

select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Choice.java
martin-g: I don't remember having such problem in my Scala apps.
This would be a major issue in Scala if it is true.
I will work on a Scala app in the coming days and I'll verify this.
Having this method is not a big deal but I'd prefer to remove it. Or at least document why it has to be there.

reiern70: Maybe it is something I have to set at compiler level? As said this is not a stopper for me because I can always roll my own java class with  onInitialize() overridden

solomax:
will wait for the improvements, then merge :)

reiern70:
Let me  see  if tomorrow I find some time to clean up code

solomax:
Merged
Thanks for the contribution!

reiern70:
Thanks to you!

On Wed, Jan 13, 2016 at 1:14 PM, Maxim Solodovnik <notifications@github.com>
wrote:

> Merged
> Thanks for the contribution!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/wicketstuff/core/pull/443#issuecomment-171274317>.
>



-- 
Regards - Ernesto Reinaldo Barreiro

PR #445, opened 2016-01-01
Osgi bundle support for input-events, tinymce4, and restannotations projects

iamrakesh:
Updated maven packaging type from 'jar' to 'bundle' for below given projects.
1. Input-events
2. tinymce4
3. restannotations & restannotations-json


martin-g:
Thank you!

PR #448, opened 2016-01-10
Jamon API integration.

rene-d-menoto:
Jamon is a widely used API to monitor application performance.
In case that webapp provided by Jamon is unusable an integration
in wicket app could help.

This contribution shall provide all the stuff required to intergrate
Jamon easily.


jamon-parent/jamon-examples/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.wicketstuff</groupId>
+		<artifactId>wicketstuff-jamon-parent</artifactId>
+		<version>7.2.0-SNAPSHOT</version>
+	</parent>
+	<artifactId>jamon-examples</artifactId>
+	<name>Jamon examples</name>
+	<description>Example usage of Wicketstuff Jamon</description>
+
```
martin-g: Please add the properties for disabling the deployment, javadoc, etc.
Consult with the other -examples/pom.xml.

jamon-parent/jamon/src/main/java/org/wicketstuff/jamon/component/EmptyMarkupContainer.java
```diff
@@ -0,0 +1,44 @@
...
+	/**
+	 * Construct. Will use the given id as {@link #setMarkupId(String) markupId} and set the
+	 * {@link #setOutputMarkupId(boolean) outputMarkupId} flag to <code>true</code>.
+	 * 
+	 * @param id
+	 */
+	public EmptyMarkupContainer(String id)
+	{
+		super(id);
+		setMarkupId(id);
```
martin-g: Is this really needed?
There is a chance that there are two HTML elements with the same id in the DOM.

jamon-parent/jamon/src/main/java/org/wicketstuff/jamon/component/JamonAdminPage.java
```diff
@@ -0,0 +1,62 @@
...
+
+	public static final String PATH_TO_MONITOR_DETAILS = "monitorDetails";
+
+	/**
+	 * Construct.
+	 * 
+	 * @param rowsPerPage
+	 *            How many monitors per page should be rendered?
+	 */
+	public JamonAdminPage(int rowsPerPage)
```
martin-g: You can pass the number as page parameter and keep the page stateless. But maybe this is not really needed here.

jamon-parent/jamon/src/main/java/org/wicketstuff/jamon/component/JamonMonitorDetailsPanel.java
```diff
@@ -0,0 +1,125 @@
...
+	 * @param id
+	 *            The id.
+	 * @param monitorLabel
+	 *            The label of the {@link Monitor} to show the details of.
+	 */
+	public JamonMonitorDetailsPanel(String id, String monitorLabel)
+	{
+		super(id);
+		setOutputMarkupId(true);
+		setMarkupId(id);
```
martin-g: as before

jamon-parent/jamon/src/main/java/org/wicketstuff/jamon/monitor/JamonRepository.java
```diff
@@ -0,0 +1,141 @@
...
+ * {@link #clear()} method to remove all {@link Monitor}s from this repository.
+ * </p>
+ * 
+ * @author lars
+ * 
+ */
+@SuppressWarnings("serial")
+public class JamonRepository implements Serializable
+{
+	private static final JamonRepository jamonRepository = new JamonRepository();
```
martin-g: This could be stored in Application's metadata instead of being a `static` field.

jamon-parent/jamon/src/main/java/org/wicketstuff/jamon/request/cycle/JamonMonitoredRequestCycle.java
```diff
@@ -0,0 +1,192 @@
...
+ * {@link #includeSourceNameInMonitorLabel} set to <code>false</code>.</li>
+ * </ul>
+ * 
+ * Any navigations from or to the {@link JamonAdminPage} is excluded from the Monitors.
+ * </p>
+ * 
+ * @author lars
+ * 
+ */
+public class JamonMonitoredRequestCycle extends RequestCycle
```
martin-g: IMO there is no need of custom RequestCycle.
Better extract a POJO for its members and store it in RequestCycle's metadata. Use the callback methods in the IRequestCycleListener.

jamon-parent/jamon/src/test/java/org/wicketstuff/jamon/component/JamonAdminPageTest.java
```diff
@@ -0,0 +1,60 @@
...
+		wicketTester.startPage(new JamonAdminPage());
+		wicketTester.assertRenderedPage(JamonAdminPage.class);
+	}
+
+	@Test
+	public void shouldRenderStatisticsPageWithTwoMonitors()
+	{
+		JamonTestUtil.startThisManyMonitors(2);
+		wicketTester.startPage(new JamonAdminPage());
+		wicketTester.assertRenderedPage(JamonAdminPage.class);
```
martin-g: Since there is no assertion of the number of the monitors I think there is no need of two test methods.

rene-d-menoto:
Hi Martin, I've applied your comments. Do i need to apply a new pull request?

martin-g:
No. I will review the PR soon.
GitHub didn't send any notifications about the new commits :-/

jamon-parent/jamon-examples/src/main/java/org/wicketstuff/jamon/example/HomePage.java
```diff
@@ -0,0 +1,55 @@
...
+					new JamonAdminPage(new PageParameters().set(JamonAdminPage.PARAM_ITEMS, 100)));
+			}
+		});
+		add(new Link<Void>("toAjaxPage")
+		{
+
+			@Override
+			public void onClick()
+			{
+				setResponsePage(new AjaxPage());
```
martin-g: Prefer `BookmarkablePageLink` instead of `Link+setResponsePage()` if this is the only thing it does. 
It saves few HTTP redirects.

jamon-parent/jamon/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+			<groupId>javax.servlet</groupId>
+			<artifactId>javax.servlet-api</artifactId>
+			<scope>test</scope>
+		</dependency>
+
+		<!-- LOGGING -->
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-log4j12</artifactId>
+		</dependency>
```
martin-g: Specific logging should be used only in the examples/applications.
The modules should just define dependency on `slf4j-api`.
Otherwise an application that uses Logback will have to exclude those.

PR #451, opened 2016-01-12
In method grid fix paging disabled highlight

rototor:
Wicket 7 does no longer emits an em tag for disabled links. The styling
must be changed to reflect this, as otherwise the current selected page
is no longer visible in the navigation bar.


martin-g:
Thank you!

PR #453, opened 2016-01-15
add bootstrap theme to select2

reiern70:


select2-parent/select2/pom.xml
```diff
@@ -49,6 +49,11 @@
 			<artifactId>slf4j-simple</artifactId>
 			<scope>test</scope>
 		</dependency>
+        <dependency>
+            <groupId>de.agilecoders.wicket.webjars</groupId>
+            <artifactId>wicket-webjars</artifactId>
+            <version>0.5.3</version>
+        </dependency>
```
solomax: whitespaces :(

reiern70: I know :-( See my apologies on wicket list ;-) I should fix my settings.

select2-parent/select2/src/main/java/org/wicketstuff/select2/Settings.java
```diff
@@ -53,7 +55,6 @@
 	private String initSelection;  //TODO this will not work
 	private String query;
 	private String width;
-	private String theme;
```
solomax: I'm afraid this change breaks original meaning of 'theme' described here: https://select2.github.io/examples.html#themes-templating-responsive-design

reiern70: I think it wont: there is still a setter receiving an String (the name of the theme). To the end user noting changes.

reiern70: Disregard my comment: I need some time to digest info there.

reiern70: I still fail to see how it will break it the only thing I think we don't cover is

https://select2.github.io/examples.html#templating

solomax: Actually 'templating' from your last comment is supported by 'templateResult' settings (we are using it in our project)
You are creating 'fake' empty theme if only name was provided, my mistake, comment too early
Could you please update spaces and I'll merge the change

solomax:
I'll merge PR, then manually fix white spaces ...

reiern70:
@solomax,

Thansk!

PR #454, opened 2016-01-18
Fix #429: now it's possible to clear input current value

### LINKED ISSUE #429: It's not possible to clear input in wicketstuff-select2 without touching a model
Please find a project which demonstrate this bug here: https://github.com/PhantomYdn/wicket-bugs/tree/master/wicketstuff-select2

If you will try to clear input: it will be replaced by initial value.

PhantomYdn:


select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java
```diff
@@ -465,4 +465,12 @@ public void onComponentTagBody(final MarkupStream markupStream, final ComponentT
...
 		replaceComponentTagBody(markupStream, openTag, buffer);
 	}
+	
+	/**
+	 * Empty input is acceptable
+	 */
+	@Override
+	public boolean isInputNullable() {
+		return true;
+	}
```
solomax: I believe this shouldn't be the common option
Everyone who uses the class can override the method ....

PhantomYdn: But that is pretty natural for this type of component. So I recommend to have that by default and override if other behavior is needed: honestly, I don't see reasons for other behavior in real life:)

solomax: For some reason https://github.com/apache/wicket/blob/be18152c3d548af9e747d32385aa1f03f113cdbd/wicket-core/src/main/java/org/apache/wicket/markup/html/form/AbstractSingleSelectChoice.java#L45 has value "false" by default
I believe it will be good idea to have "false" here
@martin-g WDYT ?

PhantomYdn: @solomax Imho, it's not related to mentioned by you link. isInputNullable was overriden in AbstractTextComponent which was used as base class for AbstractSelect2Choice where, I think, it should be overriden one more time. And btw, In AbstractSingleSelectChoice method isInputNullable returns true, as it's done by this PR
@martin-g please advise what is correct.

solomax: Since latest Select2 requires <select> tag, I guess parent class can be changed
but let's wait for @martin-g  comment :)

martin-g: I think empty input is OK. I'd expect in this case to return all possible options.

How this fixes the problem with the clearing of the input?

PhantomYdn: @martin-g , if input is empty I would like to expect empty list:) That was a problem.
This fix has been tested: if you remove previously entered tags: at the end instead of showing everything it's showing empty. I think it's pretty natural: other wise it leads to bugs. Let me show real example:
1) Go to this page: http://demo.orienteer.org/class/MyClass  (login as admin/admin)
2) Click Edit
3) Try to remove ALL super classes.

PhantomYdn: Guys, any update on this? @martin-g , @solomax ?

solomax: I'll try to check this version this weekend
Can merge it in case our current application will remain stable :)

martin-g: :+1:

PhantomYdn: Thank you! Will wait.

solomax: Everything seems to work as expected
I'm going to merge it

PhantomYdn:
Thank you!

PR #456, opened 2016-01-23
Small optimizations due to martin-g comments

rene-d-menoto:
Thanks Martin for your support. I adapted your approval. Would be nice if you could integrate them.

Thanks
René


martin-g:
Thanks!

PR #457, opened 2016-01-23
Wicket 6.x

rene-d-menoto:
I'm not quite sure if this is the way to go. I would like to provide a working wicket-6 solution. I pushed the version to wicket-6.x branch. It works with jdk6u45


martin-g:
It is not clear from your description what you want to do.
But this PR has way too many unrelated commits.
I'm pretty sure this is not what you wanted but misusage of Git.

Issue #465, opened 2016-02-18
Upgrade TinyMCE 4.1.7 to the latest 4.3.3 version


simokivimaki:

martin-g:
Pull Request please!

bitstorm:
Hi,

can we close this issue?

martin-g:
Yes, the PR has been merged.

PR #466, opened 2016-02-18
Upgrade TinyMCE 4.1.7 to the latest 4.3.4 version

simokivimaki:
Issue #465


martin-g:
Thank you for the contribution!
The upgrade will be part of 7.3.0.

simokivimaki:
Thank you. I noticed that "imageupload" plugin is not included in the TinyMCE distribution but is used by ImageUploadPanel. Also some one might need compat3x plugin, so I'll add that too.

Can I push to this same pull request or make a new one?

martin-g:
This is already merged so a new will be needed.

PR #472, opened 2016-02-27
log4j dependency as test, replace commons-logging with jcl-over-slf4j

ceefour:
* log4j is only used for test
* since we use SLF4J, replace commons-logging with jcl-over-slf4j


ceefour:
replaced with #473

PR #486, opened 2016-04-25
[offline-mode] Minor improvements

martin-g:
Store the cacheable entries in the application metadata.
Use eager singletons instead of synchronizing (on the class!) for each #getInstance() call.
Add namespace to the mount paths. Used '-' instead of '/' because otherwise the url resolving breaks.
Use ResourceReference#getDependencies() to make sure that ServiceWorker is always used with ServiceWorkerRegistration.
Rename OfflineCache#load() to #renderHead() to be in sync with all other similar methods.

### DIFF
--- a/offline-mode-parent/offline-mode/src/main/java/org/wicketstuff/offline/mode/ServiceWorker.java
+++ b/offline-mode-parent/offline-mode/src/main/java/org/wicketstuff/offline/mode/ServiceWorker.java
@@ -23,6 +23,8 @@ import org.apache.wicket.Page;
 import org.apache.wicket.WicketRuntimeException;
 import org.apache.wicket.ajax.json.JSONArray;
 import org.apache.wicket.ajax.json.JSONObject;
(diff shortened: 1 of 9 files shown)


martin-g:
@klopfdreh Ping!

offline-mode-parent/offline-mode/src/main/java/org/wicketstuff/offline/mode/ServiceWorker.java
```diff
@@ -148,7 +150,7 @@ else if (cacheObject instanceof IRequestHandler)
 						}
 						else if (urlFor.toString().startsWith("."))
 						{
-							urlFor = urlFor.toString().substring(0 + 1);
+							urlFor = urlFor.toString().substring(1);
```
klopfdreh: Don't know what I was thinking by typing this. 😄

klopfdreh:
👍 - great changes!

Issue #482, opened 2016-04-07
WicketPortlet does not "portletify" the redirect url correctly 


kkaravitis:
WicketPortlet checks the response state after the action has been processed by Wicket for the presence of a redirect URL, and if present, 'portletifies' the URL.
 If the URL is a redirect to within the scope of this portlet, leaves it to be handled in  a subsequent render call, or if not, sends the redirect to the client.
 The recorded url is then used in by wicket in the subsequent 'VIEW' requests by the portal, to render the correct Page.
The way that WicketPortlet fixes the above URL in fixWicketUrl method is incorrect. The fixed url must be the wicket filter path concatenated with the URL produced from wicket framework. 
For example if the URL equals "./wicket/page?12" and the filter path is "/myWicketPortlet/" the fixed url must be "/myWicketPortlet/wicket/page?12"

kkaravitis:
I reopened this issue because in websphere portal there are still problems when the action occurs inside a bookmarkable page.

kkaravitis:
After a better study on apache wicket request life cycle, I concluded that this issue should never be opened. 
The problem of the wrong redirect url is a result of the fact that when the RequestCycle creates a UrlRenderer instance it sets as base url the client url of the current processing org.apache.wicket.request.Request instance.

```
public UrlRenderer(final Request request)
{
       this.request = request;
       baseUrl = request.getClientUrl();
}
```
In a web application the above approach is correct as it is explained in the comments of the method
org.apache.wicket.request.Request#getClientUrl() which I repeat here.

```
/**
	 * Returns the url against which the client, usually the browser, will resolve relative urls in
	 * the rendered markup. If the client is a browser this is the url in the browser's address bar.
	 * 
	 * <p>
	 * Under normal circumstances the client and request ({@link #getUrl()}) urls are the same.
	 * Handling an Ajax request, however, is a good example of when they may be different.
	 * Technologies such as XHR are free to request whatever url they wish <strong>without
	 * modifying</strong> the client url; however, any produced urls will be evaluated by the client
	 * against the client url - not against the request url.
	 * </p>
	 * <p>
	 * Lets take a simple example: <br>
	 * 
	 * Suppose we are on a client detail page. This page contains an Ajax link which opens a list of
	 * client's orders. Each order has a link that goes to the order detail page.
	 * 
	 * The client detail page is located at
	 * 
	 * <pre>
	 * /detail/customer/15
	 * </pre>
	 * 
	 * and the order detail page is located at
	 * 
[...]

PR #494, opened 2016-05-11
fix for reopened issue #482

kkaravitis:


portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
```diff
@@ -652,7 +653,21 @@ protected String fixWicketUrl(final String requestUrl, final String url) {
 		if ((url != null) && (requestUrl != null) && (!ABSOLUTE_URI_PATTERN.matcher(url).matches())) {
 			try {
 				if (!requestUrl.startsWith("http")) {
-					return new URL(new URL("http:" + wicketFilterPath), url).toString().substring(5);
+					URL fixedUrl = new URL("http:" + url);
+					String query = fixedUrl.getQuery();
+					if (query != null) {
+						for (String queryParamValuePair : query.split("&")) {
+							String wuViewParam = "_wuview=";
```
martin-g: This could be out of the loop.

kkaravitis: You are right, I will fix it as soon as possible.
Στις 11 Μαϊ 2016 10:39 μ.μ., ο χρήστης "Martin Grigorov" <
notifications@github.com> έγραψε:

PR #509, opened 2016-06-02
Fix the delete action when there are two editable grids on one page.

miguelpayet:
When a page holds two editable grids, a delete event on the second grid will be performed by the first grid. First and second positions are in the order in which they were added to the page. This happens because the delete action is propagated as a message, but this message does not hold the object that receives the event. Thus the first EditableDataTable that receives the message will process it. I have added the receiving EditableDataTable objevt into the message payload and added a check for it  in the event processing.

Changes:

- GridOperationData.java: Added a field to the GridOperationData payload to hold the EditableDataTable that receives the onClick event
- EditableDataTable.java: Added a check in the onEvent method in EditableDataTable so that each object will only process its own events.
- EditableGridActionsPanel: Added the receiving datatable object to the payload in EditableGridActionsPanel

### DIFF
--- a/editable-grid-parent/editable-grid/src/main/java/org/wicketstuff/egrid/column/EditableGridActionsPanel.java
+++ b/editable-grid-parent/editable-grid/src/main/java/org/wicketstuff/egrid/column/EditableGridActionsPanel.java
@@ -100,9 +100,10 @@ public abstract class EditableGridActionsPanel<T> extends Panel
 			@Override
 			public void onClick(AjaxRequestTarget target)
 			{
+				EditableDataTable eventTarget = rowItem.findParent(EditableDataTable.class);
 				send(getPage(), Broadcast.BREADTH, new GridOperationData<T>(OperationType.DELETE,
-					(T)rowItem.getDefaultModelObject()));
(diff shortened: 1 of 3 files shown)


editable-grid-parent/editable-grid/src/main/java/org/wicketstuff/egrid/column/EditableGridActionsPanel.java
```diff
@@ -100,9 +100,10 @@ protected void updateAjaxAttributes(AjaxRequestAttributes attributes)
 			@Override
 			public void onClick(AjaxRequestTarget target)
 			{
+				EditableDataTable eventTarget = rowItem.findParent(EditableDataTable.class);
 				send(getPage(), Broadcast.BREADTH, new GridOperationData<T>(OperationType.DELETE,
```
martin-g: IMO, broadcasting the event to the page is sub-optimal. Using the EditableDataTable should be enough.

martin-g:
Backported it to wicket-6.x branch too!

martin-g:
Thank you for the contribution, @miguelpayet !

PR #510, opened 2016-06-06
Google Charts: Text and Value marker support

ahaeber:
Add support for Google Charts Text and Value markers as an additional
marker type.

For reference see
<https://developers.google.com/chart/image/docs/gallery/line_charts#text-and--data-value-markers-chm-bar-line------radar-scatter>


martin-g:
Hi!
Thank you for your contribution but branch `core-1.4.x` is not maintained anymore!
I'll merge it anyway but most probably we won't make any new releases from this branch in the future.

ahaeber:
Thanks :) Should I make a similar PR to add this functionality to master?

martin-g:
Yes, please!
We use `master` for WicketStuff 7.x and `wicket-6.x` for WicketStuff 6.x.
You could create a PR only for one of those. We will cherry-pick it to the other.

PR #511, opened 2016-06-06
Ignore IntelliJ project files and folders

ahaeber:


martin-g:
Hi!
Thank you for your contribution but branch `core-1.4.x` is not maintained anymore!
I'll merge it anyway but most probably we won't make any new releases from this branch in the future.

PR #512, opened 2016-06-06
Google Charts: Text and Value marker support

ahaeber:
Add support for Google Charts Text and Value markers as an additional
marker type.

For reference see
https://developers.google.com/chart/image/docs/gallery/line_charts#text-and--data-value-markers-chm-bar-line------radar-scatter

Signed-off-by: Andreas Häber <andreas.haber@intele.com>


martin-g:
Thanks!
I've cherry-picked it to wicket-6.x branch too!

PR #516, opened 2016-06-10
Google Charts: Add data scaling parameter support

ahaeber:
Adds support for specifying data scaling.

Reference: https://developers.google.com/chart/image/docs/data_formats

Signed-off-by: Andreas Häber <andreas.haber@intele.com>


ahaeber:
Support for another missing parameter in googlecharts.

PR #519, opened 2016-06-26
Separate monitoring data from components. Fixes #518

### LINKED ISSUE #518: Use JamonAdminPage without logging requests(Maybe make JamonRepository static, or use IInitializer)
JamonRepository is a wrapper around MonitorFactory which is static. Currently JamonRepository holds no state however it is set as meta data on the Application class. This is not necessary since it's really a utils type wrapper over MonitorFactory.

If there is a reason that you want to make it be an instance variable then I suggest creating an IInitializer to set it as metadata in the Application.

The justification for this change is for people(like me) that don't want to log requests by using the JamonAwareRequestCycleListener but do want to use the wicket components to view jamon statistics in other parts of the application. By forcing the use of the JamonAwareRequestCycleListener this is not possible because it is the one that sets the meta data in the Application.

rene-d-menoto:
The repository providing the monitoring data is required by components.
But coupling the repository to application makes it hard to reuse
components without repository.

The repository is still required for components but can be separated
from monitoring statement creation. So you shall be able to stop
monitoring components but use the monitoring components e.g. as a
performance monitor.


martin-g:
Thanks!

PR #525, opened 2016-07-07
ofchart wicket 6 fix

philipreimer:
replace ofchart dependency because old dependency's repository http://jofc2.googlecode.com/svn/repository/snapshots/ is not available anymore


martin-g:
Does that need to be ported to `master` (Wicket 7.x) ?

philipreimer:
I don't think so. `master` already uses

```
		<!-- JOFC2 (open flash chart for java) -->
		<dependency>
			<groupId>ro.nextreports</groupId>
			<artifactId>jofc2</artifactId>
			<version>${jofc2.version}</version>
		</dependency>
```

martin-g:
Perfect!
Thank you!

PR #527, opened 2016-07-08
add missing IInitializer files so that MonitoringRepository gets regi…

philipreimer:
…stered with Application


martin-g:
Thanks!

Issue #531, opened 2016-08-03
Wicket portlet do not work at all when it is into a served over SSL (HTTPS) portal page 


kkaravitis:
https://github.com/wicketstuff/core/blob/master/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java#L654

kkaravitis:
Apache wicket offers to the wicket users the option to choose which web pages of their web application will be secured by using the HTTPS mounting, a special request mapper called HttpsMapper and by using the RequireHttps annotation with those pages they want to serve over SSL (([https://ci.apache.org/projects/wicket/guide/7.x/guide/single.html#security_3])).

The above is a perfect feature for web applications but in the "portlets world" the portal page is responsible to protect it's portlets over SSL. 

The major bug of the wicket portlet bridge is that it does not take into account the case that the portlet may be inside a protected (HTTPS) portal page and handles all the https requests to the wicket application/portlet as unsecured (http) requests.

PR #532, opened 2016-08-09
Fixes for making the wicket portlet bridge HTTP/HTTPS scheme aware

kkaravitis:
This pull request fixes the #531 issue.

The PortletRequestMapper is wrapped over an HttpsMapper and the desired scheme is always the scheme of the portlet request. 
In other words the portal page request drives the desired scheme of the wicket portlet's page. If a portal page is being served over SSL then the desired scheme of the wicket page will be the https and if the same portlet exists into an other unsecured page then the desired scheme will be the http.


martin-g:
I miss BitBucket's feature to ignore whitespace changes.

martin-g:
Actually there is a way!
https://github.com/wicketstuff/core/pull/532/files?w=1

martin-g:
Looks good! :+1:

PR #538, opened 2016-09-05
[select2] Automatically include i18n resource file

tgoetz:
... based on configured (or detected) language

### DIFF
--- a/select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2ResourcesBehavior.java
+++ b/select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2ResourcesBehavior.java
@@ -12,16 +12,12 @@
  */
 package org.wicketstuff.select2;
 
-import java.net.URL;
-
 import org.apache.wicket.Application;
 import org.apache.wicket.Component;
-import org.apache.wicket.Session;
 import org.apache.wicket.behavior.Behavior;
 import org.apache.wicket.markup.head.CssHeaderItem;
 import org.apache.wicket.markup.head.IHeaderResponse;
 import org.apache.wicket.markup.head.JavaScriptHeaderItem;
-import org.apache.wicket.request.resource.JavaScriptResourceReference;
 
 /**
  * Adds various resources needed by Select2 such as JavaScript and CSS. Which resources are added is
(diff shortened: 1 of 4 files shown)


select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2ResourcesBehavior.java
```diff
@@ -60,20 +56,5 @@ public void renderHead(Component component, IHeaderResponse response)
 		{
 			response.render(CssHeaderItem.forReference(settings.getCssReference()));
 		}
-		//i18n
-		try
-		{
-			String name = String.format("res/js/i18n/%s.js", Session.get().getLocale().toLanguageTag());
-			URL lang = getClass().getResource(name);
-			if (lang != null)
```
solomax: Previously there was a check: if i18n file for given locale exists, currently such check is missing ....

tgoetz: Yes, but do we really need such a ckeck?

1. It adds overhead to the (guessed) more than ~90% where the configured language is correct (and existing)
2. There is a fallback to "en" anyway
3. You'll see a 404 request if you configured a non-existing language
4. User with a browser that uses an unsopported language will *always* see a warning in the logfile ...

tgoetz: Ok, thinking again I got to the point that I might be a good idea to have such a check, so I added one (see latest commit).

select2-parent/select2/src/main/java/org/wicketstuff/select2/Settings.java
```diff
@@ -42,10 +44,13 @@
...
 	}
 
+	// language code (e.g. "en", "de", "fr", ...)
+	private String language;
+
 	private Integer minimumInputLength, minimumResultsForSearch;
 	private Integer maximumSelectionLength;
-	private Object placeholder;
-	private boolean allowClear;
+	private Object placeholder; // ok
```
martin-g: What is `// ok` ?

tgoetz: Oops, shouldn't have made it into the commit, was about to check which settings/options are still valid with select2 4.x. Will remove this in a minute ...

solomax: Great :)
I'm ready to merge :)

solomax:
Thanks :)

PR #553, opened 2016-10-24
fix missing upload.min.js needed for production mode

### LINKED ISSUE #551: Image Upload plugin won't work in Deployment Mode
Image upload plugin doesn't have minified version of JS, That needed for Deployment mode.

kulik:
fixed #551


bitstorm:
than you!

martin-g:
I guess this fix should be backported to wicket-7.x too.

bitstorm:
yep! I will do it

PR #558, opened 2016-10-31
Fix the save action in an EditableGrid when there are more than 1 editable grids on the page

miguelpayet:
Added a check in EditableDataTable.java to make sure the parent table of the affected grid row is the one that processes the message.

When a page holds two editable grids, the save event on a row on the second grid will be performed by the first grid. First and second positions are in the order in which they were added to the page. This happens because the save event is processed by the first EditableDataTable object that receives the message.


martin-g:
Thank you!
I've backported it to 7.x and 6.x. Will be part of 6.26.0, 7.6.0 and 8.0.0-M3.

PR #560, opened 2016-11-09
wicketstuff-portlet: fix for issue #559

kkaravitis:
When a redirection to other wicket page occurs during a resource request, e.g a redirection to the InternalErrorPage because of an Exception during a download request, the redirect url is not a valid render url and the portal cannot redirect to the other wicket page.
This pull request is a fix for the above problem.

### DIFF
--- a/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/PortletRequestMapper.java
+++ b/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/PortletRequestMapper.java
@@ -20,6 +20,7 @@ import java.util.Iterator;
 import java.util.Map;
 
 import javax.portlet.MimeResponse;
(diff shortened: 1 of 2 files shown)


bitstorm:
hi @kkaravitis ,
should we port this fix also into master branch?

kkaravitis:
Hello Andrea,
Yes we should, but I think that there are and other fixes that have not been added to the master branch.
I will check the branches.

bitstorm:
ok thank you! 🙇 🙇

kkaravitis:
The following pull requests were merged for wicket-7.x branch only, so we have to make the changes and into the master. 

#547
#548
#549
#550
#552
#556
#560

kkaravitis:
Sorry I was wrong. 
The pull requests are the following

#552
#556
#560

bitstorm:
thank you!

Issue #555, opened 2016-10-27
wicketstuff-portlet: Action, resource and render urls are not rendered right in some cases on websphere portal. 


kkaravitis:
This bug happens because the UrlRenderer tries to change the portlet url in its renderUrl method. The portlet url is being produced by the portlet container, so this url is not a wicket url and must not be processed by the UrlRenderer but must be returned as it is.

PR #561, opened 2016-11-09
wicketstuff-portlet: synchronize master with latest fixes on wicket-7.x branch

kkaravitis:
- Remove the action-scoped request attributes support usage and use
render parameters for transferring the wicket redirect url from action
phase to render phase
- fix for issue #555, the url modification must be prevented if the url
is a portlet one.
- fix for issue #559, when an exception occurs during a resource request
(not ajax), the portlet does not redirect to its InternalErrorPage page.


kkaravitis:
@bitstorm 
This pull request synchronizes the master with the latest fixes!

PR #564, opened 2016-11-29
Fix stop behavior

andrel:
Hi!

I recently adapted [raphw/wicket-async-task](https://github.com/raphw/wicket-async-task) so that it works with Wicket 8, but then learned that it is abandoned in favor of async-task in wicketstuff/core.

I see from your previous commits that you also had a problem with randomly failing tests. However, we came up with different solutions. Your solution check if `refreshBehavior` is bound before stopping it. My solution doesn't attempt to stop it on the `#onSubmit` event. (Because I assume it will be stopped by an `#refresh`-event anyway.)

I'm no Wicket expert, so I would love if you would look at my solution and see if it's wrong.


martin-g:
@bitstorm Do you have time to take a look here? You dealt with this class recently.

bitstorm:
better late than never :-). I will take a look as soon as possible!

PR #566, opened 2016-12-05
WICKET-6287 Switch from json.org to open-json

bitstorm:


bitstorm:
@solomax 
Hi, I see you worked to fix branch 7.x and 6.x after we switched to open-json, so you could help reviewing this pr for master branch.

Thanks!

solomax:
Hello @bitstorm, I don't really understand why you are changing 
`org.apache.wicket.ajax.json` with `org.json` as part of this fix?

As I understood from discussion regarding this topic `org.apache.wicket.ajax.json` will internally use openjson, so there is no need in these changes ....

klopfdreh:
Hi @solomax ,

for Apache Wicket 6.x / 7.x this is true. For 8.x we deleted the files out of the sources and let them be managed by open-json so that there is no code duplications and further improvements can be used directly by only updating the dependency.

solomax:
Ah, OK, missed that part
Will review the code ASAP

klopfdreh:
Hi @solomax ,

there are two other PRs for Wicket 6.x / 7.x in which I found out some important things:

* PR for 6.x: https://github.com/apache/wicket/pull/194
* PR for 7.x: https://github.com/apache/wicket/pull/195

In the PR od 7.x I mentioned that there are issues even if we switch to open-json because of some method signatures.

solomax:
+1 from me :)

PR #573, opened 2017-01-30
Wicket 7.x pdf.js additional functions 

scascarini:
Initial minor updates to add more functionality to Wicket/pdf.js integration.


martin-g:
Thank you for the contribution, @scascarini !

Just one minor request - please prepend the commit messages with `[prd.js] ` so that the changelog is easier to read. If the prefix is not there then I add it manually before the announcement, so it is not really critical but saves me some time :-)

PR #575, opened 2017-02-07
[pdf.js] Add print and autoscale functions

scascarini:


wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
```diff
@@ -34,16 +50,40 @@
...
                     $('#'+data.canvasId).closest('.pdfPanel').siblings().find('#page_num').val(pageNumber);
                 });
 				Wicket.Event.subscribe(WicketStuff.PDFJS.Topic.TOTAL_PAGES, function (jqEvent, total, data) {
 					$('#'+data.canvasId).closest('.pdfPanel').siblings().find('#page_count').text(total);
 				});
+				Wicket.Event.subscribe(WicketStuff.PDFJS.Topic.CURRENT_ZOOM, function (jqEvent, zoom, data) {
+
+				    var zoomDropDown =  $('#'+data.canvasId).closest('.pdfPanel').siblings().find('#zoom');
+
+				    $("option.custom").each(function() {
```
martin-g: Is `each()` really needed ?
`$("option.custom").remove()` should be the same, no ?

wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
```diff
@@ -34,16 +50,40 @@
...
+                        $(this).remove();
+                    });
+
+				    zoomDropDown.val(zoom);
+
+				    if (!zoomDropDown.val()) {
```
martin-g: Why not `if (!zoom)` ?

wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
martin-g: Maybe it would be better to visit all options and:
1) remove the `.custom` one
2) deselect any selected one if `zoom` is truthy.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
```diff
@@ -39,12 +42,20 @@
...
-                ctx = canvas.getContext('2d');
+                ctx = canvas.getContext('2d'),
+                abortingPrinting = false;
 
             var MIN_SCALE = 0.25;
             var MAX_SCALE = 10.0;
             var DEFAULT_SCALE_DELTA = 1.1;
+            var MAX_AUTO_SCALE = 1.25;
+
+            var container = $(".pdfPanel");
```
martin-g: This should be scoped to the panel related to the current `canvasId` because there might be more than one PdfPanels on the page.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
```diff
@@ -73,9 +89,116 @@
...
+            function printDocument() {
+                pageRendering = true;
+                abortingPrinting = false;
+                var pages = pdfDoc.numPages;
+
+                var body = document.getElementsByTagName('body')[0];
+                body.setAttribute('pdf-js-printing', true);
+
+                var pc = document.createElement('div');
+                pc.id = 'pdf-js-print-container';
```
martin-g: It is OK to use jQuery for DOM manipulation.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
```diff
@@ -73,9 +89,116 @@
...
+                    printPage(i, pages);
+                }
+            }
+
+            function cleanUp(error) {
+                if (abortingPrinting) { return; }
+
+                if (error) {
+                    abortingPrinting = true;
+                    window.alert("error printing!");
```
martin-g: Maybe it will be better to fire an event and let the application decide how to show it to the user: alert, console, custom dialog, ...

wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
scascarini: First - yes will do. 

Second - see next comment below

wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
scascarini: Because this test is checking if the value of zoom is in the existing set of options. If not then it won't be selected and it needs to be added as custom option. I can add comment if not clear.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
scascarini: Yes - will do.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
scascarini: The existing code in this file didn't use jQuery to get the canvas  so I intended to follow the same pattern (though I made a mistake with $('.pdfPanel') !). I can rewrite to use jQuery throughout if you prefer.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
scascarini: Absolutely it would. I need to add progress and success/fail notifications as it can take quite a while to render each page for printing. The alert is just temporary until I've added those early next week when I get back to this.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs-examples/src/main/java/org/wicketstuff/pdfjs/example/PdfJsDemoPage.html
martin-g: I see.

wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
martin-g: I think using jQuery would be safer.
But I personally don't care about old browsers.
Don't spend your time on this!

martin-g:
Thank you, @scascarini !

PR #577, opened 2017-02-13
[pdf.js] Add progress event, use jquery

scascarini:


wicketstuff-pdf-js-parent/wicketstuff-pdfjs/src/main/java/org/wicketstuff/pdfjs/res/wicketstuff-pdf.js
```diff
@@ -153,26 +149,28 @@
                     var renderTask = page.render(renderContext);
                     // Wait for rendering to finish
                     renderTask.promise.then(function () {
-                        var body = document.getElementsByTagName('body')[0];
-                        var img = document.createElement('img');
-                        img.style.width = img.width = viewport.width;
-                        img.style.height = img.height = viewport.height;
+                        var img = $('<img/>')[0];
```
martin-g: No need to unwrap the jQuery object and then wrap it immediately.

PR #578, opened 2017-02-17
Deprecate wicketstuff lambda interfaces/models and add lambda components/behaviors

ageery:
Deprecated the lambda-related functionality that is now implemented in Wicket itself.

Further, as per a discussion on the wicket-dev mailing list [1], it was suggested that components/behaviors with built in lambda functionality be implemented separately instead of adding the functionality directly to the components/behaviors in Wicket itself.   

It was further argued that using static factory methods to create "lambda-enabled" components and behaviors was not very flexible because the class could not be overridden.

This is a (hopefully) straightforward implementation (sub-classing) of adding lambda handlers to components and behaviors that have onXXX methods.  The implementation takes a lambda as a constructor parameter or setter, and uses the lambda to implement the onXXX methods while marking the onXXX methods as final.

Since these are "regular" classes as opposed to static factory methods, it is possible to (anonymously) sub-class them to provide other functionality (e.g., overriding onConfigure).  There should be a class that can be used to accomplish the functionality for each static factory method in org.apache.wicket.lambda.Lambdas.

The names of these "lambda-enabled" classes are exactly the same as the "regular" classes in Wicket itself (e.g., AjaxLink).  The difference is the package the classes exist in -- s/org.apache.wicket/org.wicketstuff.lambda/.  This should make it easy to find/replace one with the other.

[1] http://mail-archives.apache.org/mod_mbox/wicket-dev/201702.mbox/%3C839fe7c0-8e79-cc9d-aa51-03e01be85473@meiers.net%3E

### DIFF
--- a/lambda-parent/lambda/pom.xml
+++ b/lambda-parent/lambda/pom.xml
@@ -19,6 +19,10 @@
 	<dependencies>
 		<dependency>
 			<groupId>org.apache.wicket</groupId>
+			<artifactId>wicket-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.wicket</groupId>
 			<artifactId>wicket-extensions</artifactId>
 		</dependency>
 		<dependency>
--- a/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/OptionalFunction.java
+++ b/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/OptionalFunction.java
@@ -11,9 +11,12 @@ import java.util.Optional;
  *            - the type of the input to the function
  * @param <R>
  *            - the type of the result of the function
+ * @deprecated no direct replacement but use {@link org.apache.wicket.model.LambdaModel#of(org.apache.wicket.model.IModel, org.danekja.java.util.function.serializable.SerializableFunction, org.danekja.java.util.function.serializable.SerializableBiConsumer} instead           
  */
 public class OptionalFunction<T, R> implements SerializableFunction<T, R> {
 
+	private static final long serialVersionUID = 1L;
+	
 	private SerializableFunction<T, Optional<R>> opFunction;
 	private SerializableSupplier<R> defaultValueSupplier;
 
--- a/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/SerializableBiConsumer.java
+++ b/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/SerializableBiConsumer.java
@@ -10,6 +10,7 @@ import java.util.function.BiConsumer;
  *            - the type of the first argument to the operation
  * @param <U>
  *            - the type of the second argument to the operation
+ * @deprecated Use {@link org.danekja.java.util.function.serializable.SerializableBiConsumer} instead
  */
 public interface SerializableBiConsumer<T, U> extends Serializable, BiConsumer<T, U> {
 
--- a/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/SerializableBiFunction.java
+++ b/lambda-parent/lambda/src/main/java/org/wicketstuff/lambda/SerializableBiFunction.java
@@ -12,7 +12,7 @@ import java.util.function.BiFunction;
  *            - the type of the second argument to the function
  * @param <R>
  *            - the type of the result of the function
- * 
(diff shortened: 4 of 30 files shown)


solomax:
Maybe it worth to deprecate these components in wicket7, and remove in wicket8?

svenmeier:
Many thanks for taking going through all deprecations.

But we won't add all these behavior subclasses, this is not what we have agreed on on the mailing list.

ageery:
Sven, this is from the discussion on the dev mailing list from you:

"you can easily create your own subclasses that use lambdas for what you
need."

This is exactly what this pull request implements :).

Further, one of the objections to using static factory methods is that
there is no way to (anonymously) sub-class the resulting
components/behaviors.  Having lambda-related components/behaviors as
stand-alone classes alleviates this problem.

To me, this seems to be the best of both worlds.  If you want to continue
writing verbose, anonymous class onXXX handlers, continue to use the
components/behaviors in the core Wicket libraries.  If you want to write
more concise, idiomatic code, use the components/behaviors in the
wicketstuff library.

Andrew

svenmeier:
You're missing an important part in my statement: "*you* can easily create *your* own subclasses that use lambdas for what *you* need." *I* don't want these neither in Wicket nor in wicketstuff.

> Having lambda-related components/behaviors as stand-alone classes alleviates this problem.

No it does not. Your subclasses use lambdas for a single aspect of a component only: events. What about #onConfigure(), #onComponentTag(), #renderHead() ... ?
I got the impression that most devs agreed that a component is too complex to be reduced to a single lambda.

Personally I'm fine with static factories for behaviors though.

ageery:
As Martijn originally said, the problem with factory methods is that you
lose all control over sub-classing.  That is why this lambda functionality
is done using stand-alone classes rather than factory methods.  The problem
it is solving is the problem Martijn explicitly talked about in his
original message: "factory methods don't allow for additional overriding,
thus a combinatorial explosion of API".

The reason these lambdas deal with events is because that is the (in most
cases) abstract method that needs to be overridden; it's the piece you need
to add to the component to make it useful.  In 99% of the cases where I use
these components, the only methods I need to override are the onXXX methods
-- is your experience different?  This is the reason that the lambda
targets the event functionality.  And for that 1% of the cases where
additional overriding is needed, the components can be sub-classed as
needed because they're classes, not objects created by a static factory.

As Martin noted as part of the dev thread, using lambdas to specify event
handlers is a common idiom in other frames from Vaadin to JavaFX.  Java 8
has been out for three years now; we _do_ know what a good idiom for
lambdas is.  Adding lambdas for event handlers seems like an easy,
low-impact lambda usage.

svenmeier:
>using lambdas to specify event handlers is a common idiom in other frames from Vaadin to JavaFX

But Wicket doesn't have event handlers, you're taking two steps at once.

ageery:
Again, I'm thinking of the lambdas as adding that part of the component
that makes it useful.  For example, to make an AjaxLink useful, you need to
do something when the user clicks it.  I don't see that as taking two steps
at once; I see that as succinctly defining an action.

If you are for static factory methods, this pull request is exactly the
same, except that it adds the ability to sub-class, which was listed as one
of the drawbacks to static factory methods.

svenmeier:
I think we have discuss first whether we want to add event handlers to components. If yes, support for lambda events comes free.

My opinion on static factories and subclasses:
+1 static factories for behaviors
-1 static factories for components
-1 component subclasses for event handlers (in wicket core or wicketstuff)

ageery:
I thought Wicket 8 was all but wrapped up.  Is there a proposal for
introducing event handlers into components?

Are you talking about something like this:

new AjaxLink("link", new ClickListener());

where ClickListener is a functional interface and thus a lambda expression
may be used in its place?

new AjaxLink("link", (target) -> target.add(myLabel));

svenmeier:
>I thought Wicket 8 was all but wrapped up.

Me too.

>Is there a proposal for introducing event handlers into components?

Not yet, but I think this pull request is pretty close to one. The listener could be held in a behavior and the default implementation of onClick() iterates through all behaviors and looks for the interface:

    // AjaxLink#addListener(AjaxLink.ClickListener)
    new AjaxLink("link").addListener(target -> target.add(myLabel));

I'm not saying that this is the way to go, just thinking out loud. And I don't see the need to rush this into Wicket 8.0.

Closing this pull request now, since it definitely won't make it into wicketstuff.

PR #579, opened 2017-02-20
Added @deprecated and @see Javadocs for superceded functionality.

ageery:
The lambda-related component and model functionality now exists in the
Wicket libraries themselves. In addition, the Serializable lambda
interfaces are now being provided by the jdk-serializable-functional
library.


svenmeier:
Thanks!

PR #596, opened 2017-05-09
inmethodgrid: CheckBoxColumn: Fix click event name

rototor:
At the moment when using the CheckBoxColumn in Wicket 7 you get spammed
with „replace onclick with click“ messages, because you get the message
for every row in which a CheckBoxColumn is displayed…

This is already fixed for Wicket 8 in master.


svenmeier:
Thanks!

PR #600, opened 2017-06-05
Expose fit markers JS

vrozkovec:
When manipulating map with AJAX - adding or removing markers, with this change it is possible to reuse JS and add it to the target so the map fits active markers


martin-g:
@JoachimRohde Could you please review the changes ?
You can use https://github.com/wicketstuff/core/pull/600/files?w=1#diff-0 to ignore the whitespace changes.
Thank you!

Issue #602, opened 2017-07-23
UrlParametersReceivingBehavior does not find resource when subclassed


reckart:
If UrlParametersReceivingBehavior is subclassed outside its package, e.g. by creating an anonymous inner class in a page in a case where AsyncUrlFragmentAwarePage or AsyncUrlFragmentAwarePanel cannot be used, then it does not find it's `urlfragment.js` resource anymore. The reason is that the resource is looked up relative to the actual class (e.g. the anonymous inner class) and not relative to `UrlParametersReceivingBehavior` in `renderHead`:

```
	@Override
	public void renderHead(Component component, IHeaderResponse response)
	{
		// ... snip ...  
		response.render(new OnDomReadyHeaderItem(sb.toString()));
		response.render(getJS(getClass()));
	}
```

PR #603, opened 2017-07-23
#602 - UrlParametersReceivingBehavior does not find resource when subclassed

reckart:
- Look up resource relative to UrlParametersReceivingBehavior and not relative to the actual class to allow subclasses outside the wicketstuff package


reckart:
On merge, please also apply to 7.x branch.

solomax:
cherry-picked to wicket-7.x
Thanks for the contribution!

PR #612, opened 2017-09-26
exception is not logged if json payload is invalid

cserepj:
Also there is no way to signal back to the user of the API any json errors that arise during the extractMethodParameters call. In my case a misformed JSON caused an exception that was not logged and all the user saw was an http 400 status code. No exceptions in the logs, no error message in the response, nothing - while the handleException was implemented in the actual resource so that a uniform json error message structure was returned for all other error cases. I think routing any exception from extractMethodParameters through handleException method is a better way to do this.

### DIFF
--- a/wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/AbstractRestResource.java
+++ b/wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/AbstractRestResource.java
@@ -221,7 +221,7 @@ public abstract class AbstractRestResource<T extends IWebSerialDeserial> impleme
 		}
 		catch (RuntimeException e)
 		{
-			setResponseStatusCode(400);
+                        handleException(response, e);
 			return;
 		}
 


wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/AbstractRestResource.java
```diff
@@ -221,7 +221,7 @@ private void handleMethodExecution(AttributesWrapper attributesWrapper,
 		}
 		catch (RuntimeException e)
 		{
-			setResponseStatusCode(400);
```
bitstorm: Hi,

I would add a call to 'log.debug' rather than calling handleException which is meant to catch internal errors (return code 500) and not bad request errors

cserepj: In my view handleException is supposed to handle all exceptions that arise during the request and even the javadoc says the behavior to reply status 500 should be overriden.  My problem is that a bad json in request body that is supposed to be mapped to a POJO parameter of the actual method can end up in this catch block. As an API developer I want to be able to send the API caller a detailed description of what she/he is doing wrong and this catch - if not routed to the same handleException call - makes it impossible to handle such an api user error.

Doing a debug level log statement helps nothing, it's still a swallowed exceptin in most cases.

cserepj: I actually had to start  the application in debug mode, put a breakpoint in AbstractRestResource and trace through the methods to find where the 400 response code was coming from and why, get to the swallowed exception, print it to the console to get an idea what's wrong with the API. This is definitely not okay the way it's done currently. And I don't think we need a separate overridable handleExceptionDuringParameterMapping method.

bitstorm: Ok, maybe I should change the default 'handleException' removing the error code setting. Please change the PR keeping 'setResponseStatusCode(400)' and adding 'handleException' before it.

bitstorm:
Added handleException after setResponseStatusCode

PR #619, opened 2018-01-18
Added settings support for select2 behavior.

florianlacreuse:
Update AbstractSelect2Choice to use the behavior.


select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Behavior.java
```diff
@@ -13,24 +16,68 @@
  */
 public class Select2Behavior extends Behavior 
 {
-	// TODO: add settings!
 	private static final long serialVersionUID = 1L;
 
+	private Settings settings = new Settings();
+
+	public static Select2Behavior forChoice(DropDownChoice<?> choice)
```
solomax: This parameter seems to be not used ....

select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2Behavior.java
```diff
@@ -13,24 +16,68 @@
...
+	{
+		return new Select2Behavior();
+	}
+
+	public static Select2Behavior forChoice(ListMultipleChoice<?> choice)
```
solomax: This parameter seems to be not used ....

solomax:
@florianlacreuse LGTM, I'll run some tests later tonight and will merge it

florianlacreuse:
@solomax Good catch on these unused parameters! Sorry I forgot to clean it. Thanks for your help.

solomax:
@florianlacreuse everything seems to work as expected :)
Can you please add couple examples?

solomax:
Thanks a lot for the contribution! :)

PR #620, opened 2018-02-06
Debug logable widgets.

rene-d-menoto:
When we log widgets in debug mode, a NPE is thrown.
This happens when no title is provided to widget.
But it is included in toString on hashCode.
Provide a usable title.


dashboard-parent/dashboard-core/pom.xml
```diff
@@ -16,6 +16,7 @@
 
 	<properties>
 		<xstream.version>1.4.10</xstream.version>
+		<jetty.version>9.2.22.v20170606</jetty.version>
```
solomax: This is not required, defined in parent pom

dashboard-parent/dashboard-core/pom.xml
```diff
@@ -31,5 +32,33 @@
 			<artifactId>xstream</artifactId>
 			<version>${xstream.version}</version>
 		</dependency>
+
```
solomax: All this changes are not required, defined in parent pom

dashboard-parent/dashboard-core/src/main/java/org/wicketstuff/dashboard/AbstractWidget.java
```diff
@@ -31,6 +31,7 @@
 	protected WidgetLocation location;
 
 	public AbstractWidget() {
+		this.title = "";
```
solomax: I would better rewrite `hashcode` and `equals`
it uses both title and id
Any IDE can provide failsafe implementation

dashboard-parent/dashboard-core/src/test/java/org/wicketstuff/dashboard/AbstractWidgetTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private WicketTester browser;
+	
+	@Before
+	public void setup() {
+		browser = new WicketTester(new WebApp());
+	}
+	
+	@Test
+	public void can_expose_configuration() {
```
solomax: Camel case please :)

dashboard-parent/dashboard-core/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,10 @@
+log4j.appender.Stdout=org.apache.log4j.ConsoleAppender
```
solomax: `slf4j-simple` is already in charge
this file is not required

dashboard-parent/dashboard-core/pom.xml
rene-d-menoto: This was not working so far. But now i got it.

dashboard-parent/dashboard-core/src/main/java/org/wicketstuff/dashboard/AbstractWidget.java
rene-d-menoto: Added a eclipse generated version.

dashboard-parent/dashboard-core/pom.xml
rene-d-menoto: I tried to use as minimal dependencies as possible.

dashboard-parent/dashboard-core/src/test/java/org/wicketstuff/dashboard/AbstractWidgetTest.java
rene-d-menoto: Renamed

dashboard-parent/dashboard-core/src/test/resources/log4j.properties
rene-d-menoto: Removed

dashboard-parent/dashboard-core/pom.xml
solomax: Actually jetty is usually being added to the `examples` sub-project: https://github.com/wicketstuff/core/blob/master/dashboard-parent/dashboard-examples/pom.xml
This keeps main library clean and small
Are you sure this is required here?

rene-d-menoto: Do i need to implement my tests in example project? I like to provide tests that guarantee my usecase works. But then i need junit and with wicket tester i need servlet library. So far i always used jetty to provide this. Do you have another example how i can test in core library without using dependencies?

solomax: You can find example here https://github.com/wicketstuff/core/blob/master/select2-parent/select2/pom.xml#L41
I guess this way it will work and keep as small as possible :)

rene-d-menoto: Thanks. Refactored to use minimal dependencies.

solomax:
Thanks for the contribution!

Issue #621, opened 2018-02-12
Dependency conflict problem: conflicting classes exist in different JARS


PandaMonkey:
Hi, we found that org.wicketstuff:wicketstuff-closure-compiler:8.0.0 contains conflicting classes in different JARS. This means that, there are duplicate classes with the same names in different JARs and these classes have different implementations. As the JVM only load the classes present first on the classpath and shadow the other duplicate ones, so this problem brings high risk of **NoSuchMethodError** or **NoSuchMethodExceptions** at runtime.  The detail conflicting information is listed below. Please pay attention about that. Maybe shade one of them is a good choice. Hope this report can help you. Thanks!

===============projectPath->org.wicketstuff:wicketstuff-closure-compiler:8.0.0-M8@D:\ws\release\core-wicketstuff-core-8.0.0-M8\closure-compiler\pom.xml
1. Risk for jar-pair:<com.google.code.gson:gson:2.8.2:><com.google.javascript:closure-compiler:v20170910:>
   **a) Methods that only exist in com.google.code.gson:gson:2.8.2::  [com.google.code.gson:gson:2.8.2:]**
<com.google.gson.annotations.JsonAdapter: boolean nullSafe()>
<com.google.gson.Gson$1: java.lang.Double read(com.google.gson.stream.JsonReader)>
<com.google.gson.Gson$1: void write(com.google.gson.stream.JsonWriter,java.lang.Object)>
<com.google.gson.Gson$1: void <init>(com.google.gson.Gson)>
<com.google.gson.Gson$1: void write(com.google.gson.stream.JsonWriter,java.lang.Number)>
<com.google.gson.Gson$1: java.lang.Object read(com.google.gson.stream.JsonReader)>
<com.google.gson.Gson$2: java.lang.Float read(com.google.gson.stream.JsonReader)>
<com.google.gson.Gson$3: void <init>()>
<com.google.gson.Gson$3: java.lang.Number read(com.google.gson.stream.JsonReader)>
<com.google.gson.Gson$4: java.util.concurrent.atomic.AtomicLong read(com.google.gson.stream.JsonReader)>
<com.google.gson.Gson$4: void <init>(com.google.gson.TypeAdapter)>
<com.google.gson.Gson$4: void write(com.google.gson.stream.JsonWriter,java.util.concurrent.atomic.AtomicLong)>
<com.google.gson.Gson$5: void write(com.google.gson.stream.JsonWriter,java.util.concurrent.atomic.AtomicLongArray)>
[...]

PR #622, opened 2018-02-12
Pull request for issue #621

PandaMonkey:
Pull request for issue #621 "Dependency conflict problem: conflicting classes exist in different JARS"

The root cause of dependency conflict problem is that this project depends on com.google.javascript:closure-compiler:v20171203 and com.google.code.gson:gson:2.8.2, however,
com.google.javascript:closure-compiler:v20171203 is a repackaged fat JAR. By analyzing com.google.javascript:closure-compiler:v20171203, we found it contains library com.google.code.gson:gson:2.7. So that the conflicting classes occur. To solve the problem, we can harmonize the library version to com.google.code.gson:gson: 2.7, to make the world better.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -674,7 +674,7 @@
 			<dependency>
 				<groupId>com.google.code.gson</groupId>
 				<artifactId>gson</artifactId>
-				<version>2.8.2</version>
+				<version>2.7</version>
 			</dependency>
 			<dependency>
 				<groupId>com.fasterxml.jackson.core</groupId>


solomax:
Maybe it would be better to

1. update closure-compiler to most recent version
2. check which submodules might have conflict and fix dependencies of these modules?

I don't really like the idea of using old libraries ....

PandaMonkey:
@solomax @martin-g Thanks a lot for your support.
1. We have checked the newer version of closure-compiler, it also uses com.google.code.gson:gson: 2.7, so upgrading the closure-compiler cannot solve this problem.
So I have reported this issue in closure-compiler project.
See https://github.com/google/closure-compiler/issues/2815

2. Second, If we don't adjust the version of com.google.code.gson:gson, then we are also not sure which version of class are actually loaded at runtime.

I agree using old libraries is not a best idea, we can notice this dependency conflict problem, and upgrade the closure-compiler after it upgrading library com.google.code.gson:gson from 2.7 to 2.8.2.

Hope my report can help you. Thanks again.

solomax:
Hello @PandaMonkey,

What I'm trying to say: to be affected be this incompatibility wicketstuff sub-module **MUST** have dependency on both `closure-compiler` and `com.google.code.gson:gson`

The only module dependent on `closure-compiler` is `wicketstuff-closure-compiler` which is **NOT** dependent on `com.google.code.gson:gson` ....

So proposed downgrading of `com.google.code.gson:gson` version make no sense to me :(

solomax:
`mvn dependency:tree`

displays:
```
[INFO] +- com.google.javascript:closure-compiler:jar:v20171203:compile
[INFO] |  +- com.google.javascript:closure-compiler-externs:jar:v20171203:compile
[INFO] |  +- args4j:args4j:jar:2.33:compile
[INFO] |  +- com.google.errorprone:error_prone_annotations:jar:2.0.18:compile
[INFO] |  +- com.google.guava:guava:jar:20.0:compile
[INFO] |  +- com.google.protobuf:protobuf-java:jar:3.0.2:compile
[INFO] |  +- com.google.code.gson:gson:jar:2.8.2:compile
```

So it might be good idea to add `com.google.code.gson:gson` to `exclusions` of `com.google.javascript:closure-compiler:jar:v20171203` ....

solomax:
The original issue was fixed via https://github.com/wicketstuff/core/commit/81b81a26922d5f929c81c13678cd533baa5fab90
I believe this is more clean way of fixing this
Thanks for pointing this out

PR #640, opened 2018-10-24
GChart - Correct getOptionModel() and remove unused logger

tahitiangabriel:
getOptionModel() was returning the object and not the model, thus a ClassCastException was thrown.
I have also removed the unused logger.

Sorry for the CRLF, I'm using "core.autocrlf true", so it should be good, but it is not with this project...


solomax:
Hello @tahitiangabriel,
it seems your PR is almost "non-cherry-pickable" to 8.x so I'm going to perform following changes:
1) change getOptionModel()
2) remove unused Logger
3) add changes from 0627ac9

anything else need to be done here?

solomax:
@tahitiangabriel please check 8ad7edf

solomax:
Thanks for the contribution!

PR #651, opened 2019-02-11
Ajax redirect keeps ResponseState

athapostolou:
Ajax redirect now correctly keep all response state, including cookies, and is not cleared before flushed.


solomax:
I'll cherry-pick previous PR to all branches, PRs seems to be identical

Issue #650, opened 2019-02-04
wicketstuff-portlet: ajax redirect does not add cookies to browser


athapostolou:
While adding a cookie to an ajax redirect, the cookie is not set on the browser.

I have noticed that in WicketPortlet, when an ajax redirect is needed, you call `responseState.clear();`, which clears the cookie that was previously added, before the responseState is flushed. 
The portal bridge on wicket 1.4 does not clear this before flushing and therefore adds the cookie to the browser.
Is this by design, or can we change it so that ajaxRedirect also adds the cookies when wanted?

martin-g:
@kkaravitis @klopfdreh Can you help here ? Thank you!

kkaravitis:
Hello all,
I think that it is safe to keep the cookies before clearing the
responseState object in this case.
Unfortunately I do not have enough free time to make the change and test
it.
Could someone else evolve with the wicketstuff portlet project and make the
commits required?


Στις Δευ, 4 Φεβ 2019 στις 1:34 μ.μ., ο/η Martin Grigorov <
notifications@github.com> έγραψε:

athapostolou:
I could do it myself if you guys want to.
I already have tested the code for ibm portal.
I can test it for liferay also and commit it.

solomax:
PRs are always welcome :)

martin-g:
Please send a Pull Request, @athapostolou ! Thank you!

martin-g:
PR #652 has been merged to all branches!

PR #476, opened 2016-03-22
wicketstuff portlet fixies

kkaravitis:
I made changes on wicketstuff portlet and I managed to make the wicketstuff portlet bridge functional.
I tested the wicket 7.x examples as portlets succesfully on the following enviroments:
- websphere portal 8.5 
- liferay 6.1 GA3
- liferay 6.2 GA6

PR #652, opened 2019-02-11
Ajax redirect keeps ResponseState

athapostolou:
Ajax redirect now correctly keep all response state, including cookies, and is not cleared before flushed.
Related to https://github.com/wicketstuff/core/issues/650

### DIFF
--- a/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
+++ b/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
@@ -412,23 +412,7 @@ public class WicketPortlet extends GenericPortlet {
 				// processing the request
 
 				String redirectLocation = responseState.getRedirectLocation();
-				String ajaxRedirectLocation = responseState.getAjaxRedirectLocation();
-				if (ajaxRedirectLocation != null) {
-					// Ajax redirect
-					responseState.clear();
-					responseState.setDateHeader("Date", System.currentTimeMillis());
-					responseState.setDateHeader("Expires", 0);
-					responseState.setHeader("Pragma", "no-cache");
-					responseState.setHeader("Cache-Control", "no-cache, no-store");
-					//client side javascript needs the Ajax-Location header see wicket-ajax-jquery.js line 771
-					responseState.setHeader("Ajax-Location", ajaxRedirectLocation);//
-					responseState.setContentType("text/xml;charset=UTF-8");
-					responseState.getWriter().write(
-						"<ajax-response><redirect><![CDATA[" + ajaxRedirectLocation +
-							"]]></redirect></ajax-response>");
-					responseState.flushAndClose();
-				}
-				else if (redirectLocation != null) {
+				if (redirectLocation != null) {
 					// TODO: check if its redirect to wicket page (find _wu or
 					// _wuPortletMode or resourceId parameter)
 
@@ -510,7 +494,7 @@ public class WicketPortlet extends GenericPortlet {
 				}
 				else {
 					if (LOG.isDebugEnabled()) {
-						LOG.debug("ajax redirect url after inclusion:" + redirectLocation);
+						LOG.debug("ajax redirect url after inclusion:" + responseState.getAjaxRedirectLocation());
 					}
 					// write response state out to the PortletResponse
 					responseState.flushAndClose();


solomax:
From what I can see you just have dropped the code for `responseState.getAjaxRedirectLocation();`
I'm OK with it, but it looks suspicious ...

portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
```diff
@@ -510,7 +494,7 @@ else if (redirectLocation != null) {
 				}
 				else {
 					if (LOG.isDebugEnabled()) {
-						LOG.debug("ajax redirect url after inclusion:" + redirectLocation);
+						LOG.debug("ajax redirect url after inclusion:" + responseState.getAjaxRedirectLocation());
```
solomax: I would use `LOG.debug("ajax redirect url after inclusion: {}", responseState.getAjaxRedirectLocation());` here

svenmeier:
Could you enhance this pull request with some context: description, problem, reference to discussions ...

Seems strange to fix things with just deleting a chunk of code (although that happens sometimes).

martin-g:
Related issue: https://github.com/wicketstuff/core/issues/650

portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
```diff
@@ -412,23 +412,7 @@ private void processMimeResponseRequest(String wicketURL, final PortletRequest r
 				// processing the request
 
 				String redirectLocation = responseState.getRedirectLocation();
-				String ajaxRedirectLocation = responseState.getAjaxRedirectLocation();
-				if (ajaxRedirectLocation != null) {
```
martin-g: @kkaravitis suggested to preserve the cookies before clearing the state and to re-apply them after that.

athapostolou: @kkaravitis, @martin-g 
Do you think that there is a reason to clear the state at all? 
This is a redriect to any location a user wants. Inside the portal or outside it. 
A user might want to redirect to another site and add some authentication headers for example.

I understand that I found this from my cookies problem, but maybe I should rephrase it.
Unless you know of a reason for the the response state to be cleared..

portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
athapostolou: Will do this. 
Just waiting for the discussion below, to change them all together.

portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
```diff
@@ -412,23 +412,7 @@ private void processMimeResponseRequest(String wicketURL, final PortletRequest r
...
-				String ajaxRedirectLocation = responseState.getAjaxRedirectLocation();
-				if (ajaxRedirectLocation != null) {
-					// Ajax redirect
-					responseState.clear();
-					responseState.setDateHeader("Date", System.currentTimeMillis());
-					responseState.setDateHeader("Expires", 0);
-					responseState.setHeader("Pragma", "no-cache");
-					responseState.setHeader("Cache-Control", "no-cache, no-store");
-					//client side javascript needs the Ajax-Location header see wicket-ajax-jquery.js line 771
-					responseState.setHeader("Ajax-Location", ajaxRedirectLocation);//
```
kkaravitis: Client side javascript needs the Ajax-Location header in order to redirect. Please  see [wicket-ajax-jquery.js](https://github.com/wicketstuff/core/blob/master/portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/ajax/res/js/wicket-portlet-ajax.js) line 14.

kkaravitis: Also look at my very first [commit](https://github.com/wicketstuff/core/pull/476/commits/7831805b42cbb30edd33ce2f1cce1ca9f8f829df) where I fixed the redirect after ajax request behavior.

portlet-parent/wicketstuff-portlet/src/main/java/org/apache/wicket/portlet/WicketPortlet.java
kkaravitis: I insist on my first suggestion. I have doubts that the redirect occur if we remove all the logic of ajaxriderct url. Please see also my other comments on this commit.

svenmeier:
So this commit reverts special handling of Ajax redirects introduced in #476? Fine.

What will we do with ResponseState#getAjaxRedirectLocation()/AJAX_LOCATION_HEADER_NAME? This doesn't seem to be used anywhere else then (apart from your log statement).

martin-g:
> So this commit reverts special handling of Ajax redirects introduced in #476? Fine.

No, it is not fine.
The change here should preserve all headers (cookies are headers too), clear the state (i.e. reset any body, flags) and then reapply the old headers and set the specific headers for the XML response.

athapostolou:
The way I understand it, is that the ajax redirect is not part of the portal, but part of core wicket. 
The _Ajax-Location_ headers mentioned by @kkaravitis in his [commit](https://github.com/wicketstuff/core/pull/476/commits/7831805b42cbb30edd33ce2f1cce1ca9f8f829df) are already handled in [ServletWebResponse](https://github.com/apache/wicket/blob/7bef3d67c8ccc269f02e8943bf9a22c3cd5438e9/wicket-core/src/main/java/org/apache/wicket/protocol/http/servlet/ServletWebResponse.java#L271). Your original problem was probably that you cleared these headers.

My proposal allows wicket to keep handling the redirect, without the bridge interfering with it.
Since wicket does not clear the state when it redirects, I do not understand why the bridge should handle the state. I have tested this in IBM and liferay portal and I have not found any problems.

But if you feel that 
> The change here should preserve all headers (cookies are headers too), clear the state (i.e. reset any body, flags) and then reapply the old headers and set the specific headers for the XML response.
then I will make this change.

svenmeier:
@athapostolou My thoughts exactly: why should the portlet be concerned about building a proper Ajax response? I really don't have any expertise in portlet stuff, but the original 'fix' looks dubious.

kkaravitis:
@athapostolou  Your last comment made clear to me the problems that this pr solves. It is ok by me.
@martin-g I agree with @athapostolou to merge this pull request.

solomax:
OK,
I'll merge this PR tomorrow
cherry-pick it to all branches and will start 7.x release process

ganzd:
Hi, this commit in wicket-7.x destroyed ajax redirect, which now doesn't work at all.

martin-g:
I would use `!previousURL.equals(redirectLocation)` instead of `!=` but probably this is not what you ask.
Please open a new issue/PR. Wicket 7.14.0 has been just released so WicketStuff Core 7.14.0 still has a chance to include a fix.

solomax:
Just have commited 250daffaa (wicket-8.x), please check if it works for you

solomax:
@ganzd ping :)

ganzd:
Sorry, had production release, so have not time to check :) Basically i reverted to wicketstuff portlet 7.9 and it works. 
@solomax  I'm using wicket 7 in current project :)

Has anyone used crypto mapper in liferay porltet environment ? Currently working on solution.

solomax:
@ganzd the fix was trivial and this code wasn't changed for 3 years ...
Just have commited 750df2948, could you check if it works better for you?

martin-g:
@solomax Why do you think it is `3 years` ?
The code has been changed 3 months ago (Feb 17): https://github.com/wicketstuff/core/commit/169b54819be0ca918da0db14e21d8e634d04c1f3

solomax:
@martin-g the code for string comparing using `!=` is there since very beginning: https://github.com/wicketstuff/core/commit/169b54819be0ca918da0db14e21d8e634d04c1f3#diff-1079141977e66dc96cea83bbbb71ec30L437

tahitiangabriel:
@martin-g @kkaravitis @ganzd @athapostolou
I've updated to wicketstuff 7.14 from 7.11 and all my ajax redirect are broken because of this MR.
This MR should definitely be reverted.
Maybe just removing "responseState.clear();" should resolve your problem about cookies.
Keeping all the "<ajax-response><redirect>" stuff is mandatory.

By the way I'm using Liferay 7.1

martin-g:
@tahitiangabriel Did you try with latest -SNAPSHOT ? @solomax have made some improvements - https://github.com/wicketstuff/core/pull/652#issuecomment-499340569
Otherwise please send us a Pull Request that reverts it!
@ganzd also reported issues with this change and it seems @athapostolou has no interest to support his change.

tahitiangabriel:
@martin-g I used the latest snapshot that has the exact same problem : all the code about ajax redirection is removed.
What did @solomax is more kind of code cleanup and he fixed an equals that was invalid.
I will do a PR asap.

PR #656, opened 2019-02-28
Allow to customize selectize.js settings

dhs3000:
Selectize.js has a bunch of configurations, that are useful or must be tweaked. We need this to be able to enable the 'remove_button' plugin and overwrite some texts with localized strings.


martin-g:
Thank you, @dhs3000 !

martin-g:
Backported it to `wicket-7.x` and `wicket-8.x`!

dhs3000:
@martin-g Thanks for merging / backporting

PR #662, opened 2019-05-21
Fixes #641: prevent rest resource from creating a new session every time

### LINKED ISSUE #641: AbstractRestResource is not quite restful
https://github.com/wicketstuff/core/blob/d05e2d06c1bf424b6113c96eea3cfbbdb4ddc34b/wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/AbstractRestResource.java#L728

The `Session.get()` in `converter.convertToObject(value, Session.get().getLocale())` causing a session creation. If the API is called not from the browser the session will be created for each single request.

static method `toObject` is called in `org.wicketstuff.rest.utils.reflection.MethodParameter` in a few places.

andruhon:
- fix session per request issue
 - add some error handling customisations

### DIFF
--- a/wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/RestResourceLocaleLocator.java
+++ b/wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/RestResourceLocaleLocator.java
@@ -1,0 +1,58 @@
+/**
+ *  Licensed to the Apache Software Foundation (ASF) under one or more
+ *  contributor license agreements.  See the NOTICE file distributed with
+ *  this work for additional information regarding copyright ownership.
+ *  The ASF licenses this file to You under the Apache License, Version 2.0
+ *  (the "License"); you may not use this file except in compliance with
+ *  the License.  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.wicketstuff.rest.resource;
+
+import org.apache.wicket.Session;
+
+import java.util.Locale;
+
+/**
+ * A singleton which locates {@link Locale} for {@link AbstractRestResource}
+ */
+public class RestResourceLocaleLocator
+{
+
+	private static RestResourceLocaleLocator instance;
+
+	private LocaleResolver localeResolver;
+
+	public Locale getLocale()
+	{
+		return localeResolver != null ? localeResolver.getLocale() : Session.get().getLocale();
+	}
+
+	/**
+	 * Configure the locator singleton with a {@link Locale} resolver
+	 * 
+	 * @param localeResolver
+	 *            {@link LocaleResolver} can be a {@link FunctionalInterface}
+	 */
+	public void setLocaleResolver(LocaleResolver localeResolver)
+	{
+		this.localeResolver = localeResolver;
+	}
+
+	public static RestResourceLocaleLocator get()
+	{
+		if (instance == null)
+		{
+			instance = new RestResourceLocaleLocator();
+		}
+		return instance;
+	}
+
+}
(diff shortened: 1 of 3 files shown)


wicketstuff-restannotations-parent/restannotations/src/main/java/org/wicketstuff/rest/resource/RestResourceLocaleLocator.java
```diff
@@ -0,0 +1,35 @@
...
+ * A singleton which locates {@link Locale} for {@link AbstractRestResource}
+ */
+public class RestResourceLocaleLocator {
+
+	private static RestResourceLocaleLocator instance;
+
+	private LocaleResolver localeResolver;
+
+	public Locale getLocale() {
+		return localeResolver != null ? localeResolver.getLocale() : Session.get().getLocale();
```
martin-g: This could do an extra effort to prevent creating a Wicket `Session`:
```
if (Session.exists()) {
   locale = Session.get().getLocale();
} else {
   local = RequestCycle.get().getRequest().getLocale();
}
```

andruhon: What you propose is quite different from existing behavior, and could be a breaking change. The purpose of the resolver is that we can supply something like RequestCycle.get().getRequest().getLocale() or even return a static locale based on props or config instead of existing session.

Do you think we want to change the default behavior of how we  get a locale?

andruhon: Essentially this Session.get() is an issue by itself. I implemented this locale locator because a while ago one of our customers decided to start submitting a zillion requests per minute, which caused a zillion sessions, which was not very nice in terms of server resources and not quite restful. I don't remember, but I'm pretty sure tomcat has exploded.

So, maybe we should completely replace this `Session.get()` with something else.

martin-g: > which was not very nice in terms of server resources

`Session.get()` creates a **temporary** *Wicket* Session. Unless you call `session.bind()` there is no HttpSession being created!
So zillion of requests would lead to higher CPU usage due to garbage collection but not really using a lot of resources.

bitstorm: As Martin said Session.get() creates a temporary session, so se keep the resource stateless as It should be. There Is however room for improvement. We could introduce a method like 'getLocale' which can be overriden to use a custom logic to retrieve the locale.

andruhon: Thank you! Maybe my resource locator is redundant here then. I can simply write a protected getLocale method with Session.get().getLocale(), implementations can override it and use whatever locator they like. Let me check if there a particular reason for this locator to exist. What do you think?

andruhon: Back to my computer. I had a look, I do not see a particular reason for this `RestResourceLocaleLocator` to exist. If you, colleagues, do not see any value in this class I can remove it from the pull request and replace it with protected `getLocale` method.

andruhon: Ah. The place where it's called is a static method! So, the locator is, probably a solution.

andruhon:
Hi! What was decided about this PR?

solomax:
@bitstorm is the main contributor of this module,
so I believe we need one more approval :)

bitstorm:
@andruhon the PR looks good also for me but I prefer to apply it only to branch 8.x. For master branch I think we can find a simpler solution changing the current APIs. The basic idea I have is to pass add a new version of AbstractRestResource.toObject which takes as additional parameter a Locale Supplier.
In this way we don't rely on static instances to customize Locale retrieving. I can work on separate PR in the next days.

PS: for the current PR you should apply the code formatting rules you find here: https://github.com/wicketstuff/core/tree/master/config/eclipse

andruhon:
@bitstorm I changed the PR to point to wicket-8.x, also applied formatting to lines I created. 

p.s. Idea happily swallowed these formating xmls but didn't start to produce the same output as eclipse, so I had to install eclipse. Hope that formatting is now OK.

bitstorm:
@andruhon thank you!

PR #670, opened 2019-07-26
Change Geocoder to make secure (https) requests to avoid rejection by Google Fixes #669

### LINKED ISSUE #669: Geocoder in 6.x uses http instead of https - Google always rejects requests with http
For the latest 6.x wicketstuff branch:

In org.wicketstuff.gmap.geocoder.Geocoder.java

Line: 164       StringBuilder sb = new StringBuilder("http://maps.googleapis.com/maps/api/geocode/");

Needs to have https - I was thinking of making it configurable but it simply won't work with http anymore - google rejects it with:

> {
>    "error_message" : "Requests to this API must be over SSL. Load the API with \"https://\" instead of \"http://\".",
>    "results" : [],
>    "status" : "REQUEST_DENIED"
> }
> 
>

chrisco484:
Fairly simple fix. Google is very strict on non secure http requests these days. Fixes https://github.com/wicketstuff/core/issues/669


solomax:
Thanks for contribution!

PR #673, opened 2019-08-26
Update Select2 Bootstrap theme to Bootstrap 4.

florianlacreuse:
* Use Select2 BS4 theme from https://angel-vladov.github.io/select2-bootstrap-theme
* Update examples pages to use BS4
* Add examples page with Select2 components using BS4 theme

I tried to add another examples page in which Select2 components use the new BS4 theme without to have to rewrite everything from scratch, so I made a new page which extends `HomePage`. I like the idea to be able to see Select2 components with and without BS theme.


solomax:
Hello @florianlacreuse,
just run the example you have created
And surprisingly found no differences according to standard example :(
What am I missing?

florianlacreuse:
Hi @solomax,

Sorry, I'm not sure I understand what's going on. Here are some screenshots:

* **Old** - Examples page using BS2: https://i.postimg.cc/WNS7Y172/wicketstuff-select2-examples-old.png
* **New** - Examples page using BS4: https://i.postimg.cc/fwMDQMd3/wicketstuff-select2-examples-new.png
* **New** - Examples page using BS4 and BS4 theme for Select2: https://i.postimg.cc/8GHBS42s/wicketstuff-select2-examples-new-bs4.png

You can access to the new page with BS4 theme on Select2 by clicking on the button "Examples BS4 theme" in the header / jumbotron (from the home page).

Only the look and feel have changed, select2 examples are the same.

solomax:
So BS4 will now be default and the only one option?
I was able to find `/bootstrap` mounted page in examples
But I see no difference between *simple* home page and *bootstrap mounted page* what is the difference?

florianlacreuse:
**`select2-examples` module**: It use BS4 instead of BS2 to help us to build examples site / pages. There are no consequences for the people using `wicketstuff-select2`, for this part of the PR. Examples pages are just prettier and more modern (up to date). :)

-----------

**`select2` module**: `select2-boostrap.css` has been updated to the latest Select2 theme for BS4. It has not been updated for 4 years. But from now on, if you use `.setTheme(new Select2BootstrapTheme(false))` you will get this new theme to style your select2 components, but you need your app to use BS4 css style.

I didn't know if we should keep the old theme for for apps built on BS3, but I'm open to discussion. We could have two `ISelect2Theme`: `Select2Bootstrap4Theme` and `Select2Bootstrap3Theme`, if you want. But in this case, which versions of BS3? There is many versions for the Select2 BS3 theme, so which one we should take? Not an easy question if you ask me. 

-----------

About the difference between the _simple_ page (`HomePage.java`) and the _BS4 Select2 theme_ page (`HomeBootstrapPage.java`):

* In the first one, Select2 components use the default theme (like in the old examples page).

wicketstuff-select2-theme-default

* In the second one, Select2 components use the new Select2 theme for BS4. **It works great since examples pages are built with BS4 css**.

wicketstuff-select2-theme-bs4

solomax:
Thanks for the clarifications!
(I'm still unable to see any difference in 2 above pictures :(( )

florianlacreuse:
@solomax It's very subtle but the two select2 components with their dropdown don't have the same look / css style. With the default select2 theme (from select2 resources), select2 components don't match great with other [BS form components](https://getbootstrap.com/docs/4.3/components/forms/) (input, basic select, checkbox, etc.). The select2 component and its dropdown in the second screenshot is BS4-compliant, it will looks great with other BS4 css components. Again, it's subtle and all about margins, paddings, borders colors, etc.

Thanks for the merge **but** I would have liked you to approve the update of the `select2-bootstrap.css` from old BS to BS4.

If someone already use `wicketstuff-select2` + the old BS Select2 theme (`.setTheme(new Select2BootstrapTheme(false))`) in an app built with BS3 and update to the next version of `wicketstuff`, it will break the css style of its select2 components. He will have to create a custom `ISelect2Theme` to use its own `select2-bootstrap.css` (for BS3).

May be someone has an idea for a less breaking change?

Thanks again @solomax for your time and help!

solomax:
I'm using pure `select2`
So I would like to get some other opinios

p.s. May we will move from `wicket-jquery-ui` to `wicket-bootstrap` but it will be painful :(

florianlacreuse:
FWIW, I think it it shouldn't be the job of `wicketstuff-select2` to provide a Select2 Bootstrap theme, because of Bootstrap having several profoundly different versions.

You want a Select2 BS theme? Fine, use webjars or whatever to get the css resources and create your own `ISelect2Theme` and use `.setTheme(myCustomSelect2Theme)`.

In our projects, our Select2 components have the BS4 style but we don't use `Select2BootstrapTheme` from `wicketstuff-select2`, we choose the right BS Select2 theme css file for us. We don't want to rely on `wicketstuff-select2` updates.

Note: I didn't understand what you meant in your p.s.

solomax:
@florianlacreuse I believe it is OK to move to BS4 in `master` branch

Issue #655, opened 2019-02-21
Select2 statlesss throws error while adding default options


florianlacreuse:
Hi,

I just noticed that a stateless Select2 (using resource reference) with some default values (model not empty) will throws an exception during the `onComponentTagBody` since it tries to add select options for these values using the `provider` (`null` in this case).

https://github.com/wicketstuff/core/blob/a3f0e71a965c5b98ad90edeb96a8be7b806735c8/select2-parent/select2/src/main/java/org/wicketstuff/select2/AbstractSelect2Choice.java#L399-L404

To reproduce the issue, just add `new PropertyModel<Collection<Country>>(this, "countries")` as parameter here:

https://github.com/wicketstuff/core/blob/a3f0e71a965c5b98ad90edeb96a8be7b806735c8/select2-parent/select2-examples/src/main/java/org/wicketstuff/select2/HomePage.java#L192

```text
java.lang.IllegalStateException: Select2 choice component: statelessCountries does not have a ChoiceProvider set
     at org.wicketstuff.select2.AbstractSelect2Choice.getProvider(AbstractSelect2Choice.java:140)
     at org.wicketstuff.select2.AbstractSelect2Choice.addOption(AbstractSelect2Choice.java:402)
     at org.wicketstuff.select2.AbstractSelect2Choice.onComponentTagBody(AbstractSelect2Choice.java:442)
     at org.apache.wicket.markup.html.panel.DefaultMarkupSourcingStrategy.onComponentTagBody(DefaultMarkupSourcingStrategy.java:70)
     at org.apache.wicket.Component.internalRenderComponent(Component.java:2486)
     at org.apache.wicket.MarkupContainer.onRender(MarkupContainer.java:1593)
     at org.apache.wicket.Component.internalRender(Component.java:2291)
     at org.apache.wicket.Component.render(Component.java:2222)
```

martin-g:
The same is true also for stateful components. The application developer must provide a `ChoiceProvider` if a model is going to be used.

I will add an improvement to ignore the model when the component is stateless. This way the application can switch from stateless to stateful mode depending on external conditions.

florianlacreuse:
If a stateful select2 component doesn't have a `ChoiceProvider`, I thinks it's the developer's fault.

For a stateless select2 component, since it use a `JsonResourceReference` which provide the `ChoiceProvider`, it's a bit odd to set the `provider` attribute (the one in `AbstractSelect2Choice`).

[...]

PR #674, opened 2019-08-29
Select2 stateless displaying initial values

MargotPiva:
Select2Choice components always display initial values, except when Settings.setStateless(true) is used (field is always empty). This is because field construction uses getSettings().isStateless() to decide when filling initial values.

This problematic behavior can be seen in the updated examples provided by this PR. Without our fix, stateless example is empty despite providing initial values.

I think Select2Choice initial values should also displays (if a provider is available). I propose to fix this by modifying onComponentTagBody so that it can use the given provider, even if Settings.setStateless(true) is called.

With this fix, we can handle Settings.setStateless(true) + initial values selection, by providing : 

- A provider Select2 that can display initial values
- A mountPath to access the external provider (which is the actual behavior)
- A model with initial values

This pull request also updates the stateless Select2 example field. It is now filled with initial values which is consistent with the other examples.

Separately from the main subject, I saw that the .gitignore was provided with Eclipse files to ignore. I added ".factorypath" files that are also created by Eclipse.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .metadata
 .settings
 .project
+.factorypath
 target/
 .classpath
 *.i[mpw][lrs]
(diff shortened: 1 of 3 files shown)


solomax:
LGTM
@martin-g could you please take a look at this PR?

solomax:
@MargotPiva Thanks for contribution!

florianlacreuse:
Thanks for the changes and the quick merge.

@martin-g This fix the issue that @lalmeras and I were talking about in #655 (see last few messages).

martin-g:
@florianlacreuse Sending a PR is the way!

florianlacreuse:
@martin-g I totally agree with that. :) @MargotPiva did this PR as soon as we had time to do some open source contributions. We hope to be able to do even more PR for WicketStuff in the future. Have a nice day!

Issue #685, opened 2019-12-10
Select2 List of Lists


cdjost:
Hi,

I am trying to create a select2 of a List<List<TYPE>>. Because the states that I am trying to represent contains multiple entries. The ChoiceProvider has the generic <List<TYPE>>. The getIdValue Method that I need to override expects a List<TYPE> as the parameter. But the parameter provided by the AbstractSelect2Choice class has just the type TYPE. To be it seems like the component is completely flatining my maps. Is that behavior expected or is it a bug?

Greetings

`Caused by: java.lang.ClassCastException: class de.XXX.State cannot be cast to class java.util.List (de.XXX.State is in unnamed module of loader 'deployment.XXX-ear-development-work.ear' @7ab7edfd; java.util.List is in module java.base of loader 'bootstrap')
	at deployment.pritt-ear-development-work.ear.pritt-web.war//de.XXX.common.RequestStateS2Provider.getIdValue(RequestStateS2Provider.java:17)
	at deployment.pritt-ear-development-work.ear.pritt-web.war//org.wicketstuff.select2.AbstractSelect2Choice.addOption(AbstractSelect2Choice.java:402)
	at deployment.pritt-ear-development-work.ear.pritt-web.war//org.wicketstuff.select2.AbstractSelect2Choice.onComponentTagBody(AbstractSelect2Choice.java:443)
	at deployment.pritt-ear-development-work.ear.pritt-web.war//org.apache.wicket.markup.html.panel.DefaultMarkupSourcingStrategy.onComponentTagBody(DefaultMarkupSourcingStrategy.java:70)
	at deployment.pritt-ear-development-work.ear.pritt-web.war//org.apache.wicket.Component.internalRenderComponent(Component.java:2491)`

svenmeier:
I can't check it at the moment, but I doubt that your list of list is unpacked by select2.

Quickstart or refence to your code here on GitHub?

cdjost:
The code is not hosted. I will give you a quick start.

This is the method that creates the select2:
`public static void addStatusSelect2(final String wicketId, IModel<List<ERequestState>> requestStateModel,
      final List<List<ERequestState>> possibleStates,
      final FilterForm filterForm) {
    Select2Choice<List<ERequestState>> statusSelect = new Select2Choice<>(wicketId, requestStateModel, new RequestStateS2Provider(possibleStates));
    statusSelect.getSettings().setCloseOnSelect(true);
    filterForm.queue(statusSelect);
  }`

[...]

PR #686, opened 2019-12-12
Issue #685 list of lists

cdjost:


select2-parent/select2/src/main/java/org/wicketstuff/select2/Select2MultiChoice.java
```diff
@@ -12,128 +12,118 @@
...
+
 /**
  * Multi-select Select2 component. Should be attached to a {@code <select></select>} element.
  *
- * @param <T>
- *            type of choice object
+ * @param <T> type of choice object
  * @author igor
  */
-public class Select2MultiChoice<T> extends AbstractSelect2Choice<T, Collection<T>>
```
solomax: Could you please revert white-space changes?
All tabs were replaced with spaces :(

cdjost: Changed it.

svenmeier:
Many thanks @cdjost !

Issue #688, opened 2020-02-25
[Select2] Settings should honor Select2 Defaults


renoth:
The Settings are not configured with the Select2 defaults, for example:

closeOnSelect is set to false, the default setting is true however:

![image](https://user-images.githubusercontent.com/4069007/75265547-5b606400-57f1-11ea-8729-d8fc1bb62c1b.png)

solomax:
Changing this default setting will silently affect many users
So I would vote against such change (regardless the fact it is trivial)

martin-g:
I agree with @solomax that this ship has sailed.
Maybe only for 9.0.0 ?! But again I am not sure it is worth the break.

renoth:
I understand, it was very confusing when trying to understand the behavior when compared to the default demo of wicketstuff. Maybe a little warning in the Javadoc could be helpful for other users to avoid hours of searching for the reason of specific behaviors.

reiern70:
PRs are welcome.

PR #689, opened 2020-03-10
#688 Select2 Settings should honor default

renoth:


solomax:
Thanks for the contribution!

PR #698, opened 2020-07-17
Bootstrap version 3

Geel97:
Overlays (and popups) in OpenLayers3 depends on Bootstrap 3 (wicket-bootstrap-core version 4). OL3 cannot run on Bootstrap 4. I have tested that configuration with Wicket 9.


solomax:
Thanks for the contribution!

PR #713, opened 2021-11-11
Improve MAVEN build Performance

hj987654321:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1120,6 +1120,10 @@
 						<showDeprecation>true</showDeprecation>
 						<showWarnings>true</showWarnings>
 						<verbose>false</verbose>
+						<fork>true</fork>
+						<useIncrementalCompilation>true</useIncrementalCompilation>
+
+
 					</configuration>
 				</plugin>
 				<plugin>


pom.xml
```diff
@@ -1120,6 +1120,10 @@
 						<showDeprecation>true</showDeprecation>
 						<showWarnings>true</showWarnings>
 						<verbose>false</verbose>
+						<fork>true</fork>
+						<useIncrementalCompilation>true</useIncrementalCompilation>
+
+
```
martin-g, line 1126: These two empty lines could be removed!

martin-g (commented):
Could you please share what are the build times for you before and after these improvements ?
I use Maven Daemon by default for every Maven based project locally and all this comes for free.

pom.xml
martin-g: `useIncrementalCompilation` is already `true`. IMO there is no need to be explicit.

mpkorstanje:
Would recommend against merging. This user has spammed several projects with the same/similar PR. 

The PR is unlikely to make any improvement as noted here: https://github.com/apache/maven-surefire/pull/394#issuecomment-966250959

PR #749, opened 2023-04-20
bump jasperreports to 6.20.3 and explicitly depend on openpdf

ursjoss:
jasperreports in version 6.20.2 made the dependency on openpdf optional (see [pom](https://github.com/TIBCOSoftware/jasperreports/blob/302f46434bc66c04d8a4cde60a4333904ae36507/jasperreports/pom.xml#L68)), thus not bringing in openpdf as a transitive dependency to consuming projects anymore.

It seems that in how we use jasperreports via wicketstuff does need openpdf. I worked around it in my project by explicitly depending on the library (see [here](https://github.com/ursjoss/scipamato/pull/486/files)).

However, IMHO it would be convenient to have wicketstuff declare the dependency instead of the consumers.

Note that this not only adds to the maintenance cost of dependency management in wicketstuff, it also introduces a new repository, as jasperreports maintains their own version of the library, which is not present in maven central.

So please consider if this additional burden is worth the effort. I could understand if you don't want to accept that.

Thanks for evaluating it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -306,7 +306,8 @@
 		<jts.version>1.13</jts.version>
 		<gson.version>2.10.1</gson.version>
 		<wicket-bootstrap-core.version>4.0.4</wicket-bootstrap-core.version> <!-- 4.0.x is required for openlayers-3 (bootstrap3) -->
-		<jasperreports.version>6.20.1</jasperreports.version>
+		<jasperreports.version>6.20.2</jasperreports.version>
+		<openpdf.version>1.3.30.jaspersoft.2</openpdf.version>
 		<bcprov.version>1.70</bcprov.version>
 		<joda-time.version>2.12.5</joda-time.version>
 		<ehcache.version>3.10.8</ehcache.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -429,6 +430,13 @@
 		</dependency>
 	</dependencies>
 
+	<repositories>
+		<repository>
+			<id>jaspersoft-third-party</id>
```
martin-g, line 435: It is almost self-documented but I think it would be still good to add a comment here that lists the names of the dependencies requiring this repository.

martin-g (approved):

pom.xml
ursjoss: Makes a lot of sense. I added one.

solomax:
@ursjoss  Nope, this is because of `master` branch of wicket was changed
I'll try to fix this a bit later :)

solomax:
@ursjoss I believe you need to pull/rebase :)

ursjoss:
> @ursjoss I believe you need to pull/rebase :)

Thanks, @solomax, will do so soon from my other laptop...

ursjoss:
> @ursjoss I believe you need to pull/rebase :)

Thanks for fixing the build @solomax, that worked.

ursjoss:
Updated to bump to just relesed version 6.20.3.

PR #750, opened 2023-05-01
bump jasperreports to 6.20.4 and fall back to transitive dependency on openpdf

ursjoss:
With version [jasperreports-6.20.4](https://github.com/TIBCOSoftware/jasperreports/compare/6.20.3...6.20.4), Jasperreports reverted to make openpdf mandatory again - as it was before 6.20.2.

With this, we can remove the additional complexity (introduced in #749) with the extra repo and explicit dependency on openpdf.

As discussed [here](https://github.com/TIBCOSoftware/jasperreports/issues/351), this topic might come back later, potentially with jasperreports 7.0.0.

### DIFF
--- a/jasperreports-parent/jasperreports/pom.xml
+++ b/jasperreports-parent/jasperreports/pom.xml
@@ -35,10 +35,6 @@
 			<artifactId>jasperreports</artifactId>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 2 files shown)


solomax:
Thanks @ursjoss !

PR #847, opened 2024-03-18
tinymce6: fix webjars version

vrozkovec:
pom.xml includes version 6.8.3

### DIFF
--- a/tinymce6-parent/tinymce6/src/main/java/org/wicketstuff/tinymce6/settings/TinyMCESettings.java
+++ b/tinymce6-parent/tinymce6/src/main/java/org/wicketstuff/tinymce6/settings/TinyMCESettings.java
@@ -50,9 +50,9 @@ public class TinyMCESettings implements Serializable
 	private static final String NEWLINE_TAB = "\n\t";
 
 	public static final ResourceReference TINYMCE_JS_REF = new JavaScriptResourceReference(
-		TinyMceBehavior.class, "../../../webjars/tinymce/6.8.2/tinymce.js");
+		TinyMceBehavior.class, "../../../webjars/tinymce/6.8.3/tinymce.js");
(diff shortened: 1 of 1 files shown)


solomax:
@vrozkovec maybe it would be possible to use `latest`?

vrozkovec:
It would be the best but I didn't get to it yet, will check later if/how that is possible.

martin-g:
You just need to replace the version (x.y.z) with ‘latest’. wicket-webjars
has logic to resolve the version from the resources in the classpath

vrozkovec:
Shouldn't it be `current` instead? Looking here: https://github.com/martin-g/wicket-webjars/blob/31b02fcf6e4f5e6bd7a3cf8895e75dfa1a6dd787/library/src/main/java/de/agilecoders/wicket/webjars/settings/WebjarsSettings.java#L49

But maybe I'm mixing things up, this is related to wicket-bootstrap but I couldn't find any other reference to webjars in Wicket 10.x guide. Like I said, didn't really have time to dig in deeper.

solomax:
> Shouldn't it be `current` instead? Looking here: https://github.com/martin-g/wicket-webjars/blob/31b02fcf6e4f5e6bd7a3cf8895e75dfa1a6dd787/library/src/main/java/de/agilecoders/wicket/webjars/settings/WebjarsSettings.java#L49
> 

Yes, `current` is correct: https://github.com/martin-g/wicket-bootstrap/blob/wicket-10.x-bootstrap-5.x/bootstrap-extensions/src/main/java/de/agilecoders/wicket/extensions/markup/html/bootstrap/form/tempusdominus/TempusDominusBehavior.java#L52 :)

vrozkovec:
Updated

solomax:
I would say there is no need to have separate JSRef from minified library ... :)

solomax:
@vrozkovec Thanks for the contribution!

PR #873, opened 2024-04-16
Remove unused bcprov-jdk18on dependency

asvanberg:
There is no bundled `bcprov-jdk18on` dependency in Jasperreports or any of its transitive dependencies according to `mvn dependency:tree`.


solomax:
Thanks for the contribution @asvanberg !

PR #1006, opened 2024-08-20
#1005 Do not remove page from delegate when adding page

### LINKED ISSUE #1005: SessionQuotaManagingDataStore should not use remove-then-add for delegate store
`SessionQuotaManagingDataStore` currently does two steps when adding a new page:

1. Remove existing page
2. Add new page

The first step is necessary for keeping the `SessionData` that tracks size/counts of pages per session up to date. The problem is that it also propagates this removal to the delegate store. In case of a remote Redis instance and a possibly asynchronous client, this can lead to all sorts of problems.

We fixed many rare bugs in our application when we switched to a custom version of `SessionQuotaManagingDataStore` that only propagates the removal when necessary.

theigl:
Resolves #1005.

### DIFF
--- a/datastores-parent/datastore-common/src/main/java/org/wicketstuff/datastores/common/SessionQuotaManagingDataStore.java
+++ b/datastores-parent/datastore-common/src/main/java/org/wicketstuff/datastores/common/SessionQuotaManagingDataStore.java
@@ -116,7 +116,7 @@ public class SessionQuotaManagingDataStore extends DelegatingPageStore {
 		SessionData sessionData = getSessionData(context, false);
 
 		if (sessionData != null) {
-			sessionData.removePage(context, getDelegate(), serializedPage);
+			sessionData.removePage(context, getDelegate(), serializedPage, true);
 		}
 	}
 	
@@ -137,7 +137,7 @@ public class SessionQuotaManagingDataStore extends DelegatingPageStore {
 		synchronized void addPage(IPageContext context, IPageStore delegate, SerializedPage page) {
 			Args.notNull(page, "page");
 
-			removePage(context, delegate, page);
+			removePage(context, delegate, page, false);
 			
 			pages.add(new DelegatedPage(page.getPageId(), page.getData().length));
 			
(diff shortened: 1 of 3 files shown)


datastores-parent/datastore-common/src/main/java/org/wicketstuff/datastores/common/SessionQuotaManagingDataStore.java
```diff
@@ -153,10 +153,10 @@ synchronized void removeAllPages(IPageContext context, IPageStore delegate) {
 		/**
 		 * Removes a page by its identifier
 		 *
-		 * @param pageId The id of the page to remove
-		 * @return
```
solomax, line 157: `@return` is necessary here to make javadoc happy :)

datastores-parent/datastore-tests/pom.xml
```diff
@@ -76,5 +76,10 @@
 			<artifactId>websocket-jakarta-server</artifactId>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<scope>compile</scope>
```
solomax, line 82: the scope shouldn't be `compile`
It should be removed (so `test` will be used)

datastores-parent/datastore-tests/src/main/java/org/wicketstuff/datastores/common/SessionQuotaManagingDataStoreTest.java
```diff
@@ -1,6 +1,7 @@
 package org.wicketstuff.datastores.common;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.Mockito.*;
```
solomax, line 4: Could you please `expand` this asterisk import? :))

solomax (changes requested):

datastores-parent/datastore-tests/pom.xml
theigl: `compile` scope is required at the moment. See the discussion in #1008

solomax (approved):

datastores-parent/datastore-common/src/main/java/org/wicketstuff/datastores/common/SessionQuotaManagingDataStore.java
solomax: @theigl could you please address this one? :)

theigl: This is already addressed. For some reason this was not marked as outdated by GitHub.

theigl:
@solomax: Please go ahead and merge this if it is OK for you. I don't seem to have commit rights to this repository (yet).

PR #1008, opened 2024-08-20
#1007 Do not bind session as a side effect when removing pages

### LINKED ISSUE #1007: SessionQuotaManagingDataStore should not bind session when removing pages
`SessionQuotaManagingDataStore` currently binds the session as a side-effect when `removeAllPages` or `removePage` is called.

Both methods call `IPageContext.getSessionData(key, supplier)` which in the case of `DefaultPageContext`, binds the session:

https://github.com/apache/wicket/blob/a20a38c0068b4f926386bfb2c9dd46a470ff4baa/wicket-core/src/main/java/org/apache/wicket/pageStore/DefaultPageContext.java#L77-L89

The `removePage` and `removeAllPages` methods should probably do nothing if the session has not been bound yet.

theigl:
This PR passes `null` instead of a  null-supplier when pages are removed from a `SessionQuotaManagingDataStore`. This prevents binding the session as a side-effect.

Resolves #1007.

### DIFF
--- a/datastores-parent/datastore-tests/pom.xml
+++ b/datastores-parent/datastore-tests/pom.xml
@@ -76,5 +76,10 @@
 			<artifactId>websocket-jakarta-server</artifactId>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<scope>compile</scope>
+		</dependency>
 	</dependencies>
 </project>
--- a/datastores-parent/datastore-tests/pom.xml
+++ b/datastores-parent/datastore-tests/pom.xml
@@ -79,6 +79,7 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-core</artifactId>
+			<scope>test</scope>
 		</dependency>
 	</dependencies>
 </project>
--- a/datastores-parent/datastore-tests/pom.xml
+++ b/datastores-parent/datastore-tests/pom.xml
@@ -79,7 +79,6 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-core</artifactId>
-			<scope>compile</scope>
 		</dependency>
 	</dependencies>
 </project>
--- a/datastores-parent/datastore-tests/pom.xml
+++ b/datastores-parent/datastore-tests/pom.xml
@@ -79,7 +79,7 @@
 		<dependency>
 			<groupId>org.mockito</groupId>
 			<artifactId>mockito-core</artifactId>
-			<scope>test</scope>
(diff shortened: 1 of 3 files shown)


datastores-parent/datastore-tests/pom.xml
```diff
@@ -76,5 +76,10 @@
 			<artifactId>websocket-jakarta-server</artifactId>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
+			<scope>compile</scope>
```
solomax, line 82: please drop this `compile` scope (default `test` will be used in this case :)

martin-g, line 82:
```suggestion
```

martin-g (approved):

datastores-parent/datastore-tests/pom.xml
```diff
@@ -76,5 +76,9 @@
 			<artifactId>websocket-jakarta-server</artifactId>
 			<scope>compile</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.mockito</groupId>
+			<artifactId>mockito-core</artifactId>
```
martin-g, line 81:
```suggestion
			<artifactId>mockito-core</artifactId>
			<scope>test</scope>
```

datastores-parent/datastore-tests/pom.xml
martin-g: The tests are in `datastores-parent/datastore-tests/src/main/java/...` I do not remember why it has been done this way. 
The build fails now.
IMO the tests should be moved to `src/test/java/` but reverting the scope to `compile` is OKish too for this PR!

theigl: Exactly. It does not work with `test` scope at the moment.

solomax:
My bad :(
I didn't noticed this is separate `tests` module

the scope should be `compile`

sorry for the noise :(

datastores-parent/datastore-tests/pom.xml
solomax: My bad :(
I didn't noticed this is separate `tests` module

the scope should be `compile`

sorry for the noise :(

solomax (approved):

theigl:
@martin-g: Can I revert your changes to the POM so the build passes again?

martin-g:
Sure!

martin-g:
@solomax Nice commit message "Ыуеештп ьщслшещ ысщзу ещ иу сщьзшду " :-)

solomax:
> @solomax Nice commit message "Ыуеештп ьщслшещ ысщзу ещ иу сщьзшду " :-)

:)))
Sorry, forgot to switch keyboard back to English :)))

PR #1054, opened 2024-10-01
Resolve build warnings displayed at the start of the build

aldaris:
These changes should get rid of these build warnings:

```
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.wicketstuff:wicketstuff-jquery-ui-samples:war:10.3.0-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.sonatype.plugins:nexus-staging-maven-plugin is missing. @ line 150, column 12
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
[WARNING] The project org.wicketstuff:wicketstuff-core:pom:10.3.0-SNAPSHOT uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
[WARNING] The project org.wicketstuff:wicket-security-parent:pom:10.3.0-SNAPSHOT uses prerequisites which is only intended for maven-plugin projects but not for non maven-plugin projects. For such purposes you should use the maven-enforcer-plugin. See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -350,10 +350,6 @@
 		<maven-enforcer-plugin.version>3.5.0</maven-enforcer-plugin.version>
 	</properties>
 
-	<prerequisites>
-		<maven>3.5.0</maven>
-	</prerequisites>
-
 	<modules>
 		<module>offline-mode-parent</module>
 		<module>nashorn-parent</module>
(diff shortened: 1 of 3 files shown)


martin-g:
Thank you, @aldaris !
Good to see you around! :-)

PR #1225, opened 2025-06-19
Openlayers3 is updated to version 10.6.0

solomax:


keesvandieren:
Are all the examples still working in the browser? I did some research, and read that openlayers 10 has many api changes compared to openlayers 3.

solomax:
I've tested all tabs of demo app, and everything seems to work
No errors in JS console

PR #1285, opened 2025-09-22
Remove hard dependency logger implementation `jcl-over-slf4j`

asvanberg:
It should be up to the application to decide which logging implementation to use and where the various facades direct their logging.

The `<optional>` tag is not inherited from the `<dependencyManagement>` section in the parent POMs.

See https://github.com/apache/maven/issues/6974
See https://github.com/apache/maven/issues/6692


martin-g:
Thank you, @asvanberg !

Issue #1294, opened 2025-10-04
kendo-ui and jquery-ui JARs contain embedded JARs


reckart:
The kendo-ui and jquery-ui JARs contain embedded. 

I believe that should not be the case. In an OSGI-environment, those embedded JARs should probably rather be imported. And in a non-OSGI environment, it just bloats the JAR size.

PR #1295, opened 2025-10-04
Issue #1294: kendo-ui and jquery-ui JARs contain embedded JARs

reckart:
- Disable JAR embedding and generate package imports instead

### DIFF
--- a/wicketstuff-jquery-ui-parent/pom.xml
+++ b/wicketstuff-jquery-ui-parent/pom.xml
@@ -59,7 +59,6 @@
 								<instructions>
 									<Bundle-Version>${project.version}</Bundle-Version>
 									<Bundle-SymbolicName>${bundle.namespace}</Bundle-SymbolicName>
-									<Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>
 									<Export-Package>${bundle.namespace}.*;version="${project.version}"</Export-Package>
 									<Automatic-Module-Name>${auto.module.name}</Automatic-Module-Name>
 								</instructions>


reckart:
@solomax Do you have an opinion on this one?

solomax:
> @solomax Do you have an opinion on this one?

Unfortunately not :(( `bundle-plugin` is something I never work with

solomax:
@reckart shall this one be released as `10.7.1`? :))

reckart:
@reckart Not necessarily. At least not urgently. If somebody has an opinion, it would be nice to give them a moment (a few days at least) to voice that opinion before we release it. WDYT?

solomax:
sure let's wait :))) Thanks for taking care of it :))

reckart:
@sbriquet @sebfz1 Btw. if you have any insight on why you (?) embedded the dependencies in the Kendo UI JARs, please feel free to comment here as well.

sebfz1:
IIRC:
Since kendo-core is ASFv2 licensed, it's permitted to redistribute it.
Having the scripts, themes, cultures and messages directly embedded allows:
- the library to be used without having access to the internet (some
companies deny their intranet application to connect outside their walls)
- probably faster load as it's available from the same server it serves the
application.
- probably avoid CORS issues
- probably saves a certain complexity/burden as, depending on the theme,
the js dependencies to include are (was?) not similar from one theme to
another..

That's globally how I recall, but I can be wrong...

Hope this helps,
Sebastien

reckart:
This issue is not about Kendo (yes, Kendo is Apache licensed and redistributing it is absolutely fine). 

It is about JARs like the OSWAP Java HTML Sanitizer being *embedded* in the Wicket Kendo UI JAR - like a *fat JAR*. I.e. our JAR contains the oswap JAR file (and another one I forgot) inside it due to the embedding configuration that this PR here removes.

Embedding JARs in OSGi bundles is sometimes a measure to handle dependencies which are not themselves OSGI bundles. 

Anyway, sounds to me like removing the embedding configuration should not be an issue ;)

reckart:
Screenshot 2025-10-04 at 21 10 19

Here are the JARs *inside* the wicket-kendo-ui-10.7.0.jar

reckart:
Thinking about it... maybe I mistakenly added that configuration when I merged the Wicket Kendo UI codebase into this repo... need to check the history.

reckart:
No, I didn't add it. It was added in commit 57a12f28d8fc980b0e0f7ed6283769fcf4da5112 back in 2014. 

There is also a bit of discussion around this change, but none that specifically motivates why the embedding of dependencies was added: https://github.com/sebfz1/wicket-jquery-ui/pull/104

reckart:
@minikiller Do you actually need this in the Wicket Kendo UI JAR / bundling configuration?

```
<Embed-Dependency>*;scope=compile|runtime</Embed-Dependency>
```

sebfz1:
Well... simply put, I may just not have realized that it would be shaded!
:-/

reckart:
@sebfz1 np :)

@solomax Let's merge this then?

solomax:
Sure, @reckart please merge! :)

solomax:
@reckart do we need these changes in `9.x` branch?

reckart:
@solomax probably wouldn't hurt.

reckart:
@solomax I built the 9.x branch and for some reason, there don't seem to be embedded JARs on that branch despite the embedding instruction being present... *shrug*

solomax:
I guess we can also check the latest release :) (to ensure everything will be OK in next release :)))

solomax:
.... And/Or remove redundant "embedding instructions" :))

Issue #1296, opened 2025-10-04
Upgrade to latest Kendo UI


reckart:
Kendo UI is still at kendo-ui-2022.3.913 while the current version seems to be 2025.3.1002.

If we can, it would be nice if we could upgrade.

reckart:
@sbriquet @sebfz1 Did you have some script or something to facilitate the upgrade or is it manually downloading Kendo UI and dropping stuff into the proper folders?

solomax:
https://github.com/wicketstuff/core/blob/master/wicketstuff-jquery-ui-parent/wicketstuff-kendo-ui/src/main/java/org/wicketstuff/kendo/ui/resource/KendoAllResourceReference.java#L18

solomax:
seems to be not offline friendly ....
I'll try to dig some more :)

solomax:
the one located in resources seems to be taken from the CDN ....

reckart:
If you check the git history for one of the commits where the Kendo UI stuff was upgraded, you can see that it's quite a lot of files that are affected on each upgrade, including sometimes Java files (e.g. for new languages etc).

reckart:
See e.g. commit eb70bd456a654a570ac4449fe0e3cd39e98c19d5

sebfz1:
Yes, I have a script. Let me try to find it when I get home

solomax:
yes, forgot about culture resources :(

sebfz1:
Here you are:
https://github.com/sebfz1/wicket-jquery-ui/tree/wicket10.x/scripts
Normally you just need to update line 3 and 4.

It will indeed copy the theme/cultures/messages in the appropriate folder.
If there's new files (in cultures/messages - you'll detect with git status)
then the corresponding enums need to be updated.
For a new theme, I think we need to compare the download content with the
folders in wicket-kendo-ui.

Caution:
The script worked well, but if for any reason they changed the folder
structure or file name pattern, then the script will need to be updated
(they did it once, so it's not impossible...)

Hope this help,
Sebastien

sebfz1:
As Maxim noted, the CDN location in the Resource also needs to be updated.
It's the licensed version, so it's legally not possible to embed it, IIRC.

reckart:
[...]

PR #1300, opened 2025-10-06
Issue #1296: Upgrade to latest Kendo UI

reckart:
- docker-start.sh of the samples projects builds without tests (and should build without JavaDoc although that part doesn't seem to work)
- Deprecated mobile stylesheet in KendoUILibrarySettings since Kendo no longer provides separate mobile stylesheets
- Deleted most styles since Kendo doesn't supply them anymore. They still supply a ton of styles but they have different names and so we should create other modules for them
- Kendo has a new theme utils css. Not entirely sure it is needed, but created a new module for that which can be referenced as a dependency by the themes modules
- kendoMobileButtonGroup doesn't seem to exist anymore, but kendoButtonGroup does
- Upgraded resource reference in KendoAllResourceReference to latest Kendo version
- Updated Kendo UI Core JS
- Added icon font
- Added upgrade scripts for jquery and kendo-ui but due to the changed structure, at least the kendo-ui one doesn't work anymore as is and I didn't test the jquery-ui one yet

**How to test**
- Use the "docker-start.sh" in the samples project to build and deploy as a docker container. There you can test all the widgets

**To do**
* [x] Update the core JS files
* [x] Update the cultures / messages files
* [x] Fix the Kendo UI upgrade script to so we don't have to update styles/etc. manually again next time
* [x] Incorporate the remaining styles
* [x] Update icon list in `KendoIcon`

**Known to be broken**
- [x] The button example in the Kendo UI section - select button group and then click submit or ajax submit - in both cases, an NPE is generated
- [x] The icons in the buttons of the Kendo UI button example do not work. Probably the entire KendoIcon and icon font stuff needs to be aligned with the new Kendo icon font
- [x] TreeView is no longer included in the OSS Core


reckart:
@solomax I have made another round of updates on this. I'd suggest we merge this.

reiern70 (approved):
It is not possible to review such a long PR. I just trust this is Ok. Only please resolve conflicts

reiern70:
Can we also squash commits, please?

reckart:
Squash-merged.

PR #1434, opened 2026-03-24
Jasperreports are updated to 7.x

solomax:
Everything is compilable and "sort of" working

The only issue as far as I can tell: `SimplePdfPage` example look incomplete:

I can see "Before"
I can see contents of PDF
but there is no "After" and link :(((

Any help is highly appreciated! :)


reiern70:
@solomax Maybe implement EmbeddedJRReport as InlineFrame?

image

Other idea... use a lazy loading panel... So, that PDF is only asked after the page was rendered,.

solomax:
> lazy loading panel

this might do the trick. but I wonder: why main page top loading on this sub-request?

reiern70:
Could you please try an iframe? And do something similar to InlineFrame?

reiern70:
> Could you please try an iframe? And do something similar to InlineFrame?

> Note: The <object> tag was originally designed for various content types and plugins (like Java applets and Flash), many of which are no longer supported by modern browsers. Specific tags like , <audio>, <video>, and <iframe> are now the recommended alternatives for their respective media types.

solomax:
Well `<iframe>` is something being threatened as unsafe
According to what I was googled usual way to display pdf: `object`/`embed` I can try with embed

While investigating this one I've tried to check code of `Image`, `Video`, `MediaElement` etc. all of them successfully add link to some resource as an attribute of html element and nothing is broken ....

I'm frustrated :(

@martin-g maybe you can help? :))

reiern70:
> Well `<iframe>` is something being threatened as unsafe According to what I was googled usual way to display pdf: `object`/`embed` I can try with embed
> 
> While investigating this one I've tried to check code of `Image`, `Video`, `MediaElement` etc. all of them successfully add link to some resource as an attribute of html element and nothing is broken ....
> 
> I'm frustrated :(
> 

https://www.dofactory.com/html/object#:~:text=The%20element%20embeds%20external,to%20embed%20videos.

according to this iframe is recommended to embed PDF/HTML

solomax:
Hello @reiern70,

According to what I can read from the code `InlineFrame` can't be used due to it supposed to display page inside of it, not resource :(((

In case `object` is replaced with `iframe` I'm getting "partial" page

reiern70:
> Hello @reiern70,
> 
> According to what I can read from the code `InlineFrame` can't be used due to it supposed to display page inside of it, not resource :(((
> 
> In case `object` is replaced with `iframe` I'm getting "partial" page

I didn't say use it but use it as inspiration...

solomax:
@reiern70 the problem was in markup :))
the fix is as simple as `<object />` -> `<object></object>`

PR #1467, opened 2026-05-27
Fixing setup of resource name when loading i18n resources

xomit:
Checking whether a resource file exists does not work with Webjars
as previously implemented. The new check is now a simple implementation
whether the given language string is empty or not.


solomax (approved):

solomax:
Thanks a lot for the fix @xomit ! :)
I have tested and it seems everything works as expected now :)
I'm going to start the release process :)

xomit:
@solomax Thank you for the release.

PR #1477, opened 2026-06-04
add possibility to use templates based on https://github.com/BorisMoore/jsrender

reiern70:
add possibility to use templates based on https://github.com/BorisMoore/jsrender

Note: https://github.com/BorisMoore/jquery-tmpl is no longer maintained and https://github.com/BorisMoore/jsrender is the replacement for it. jquery-tmpl needs to be adapted to jQuery 4 but that will not happen.  Support for jQuery 4 for jsrender might be added soon

https://github.com/BorisMoore/jsrender/pull/377


solomax (approved):
LGTM, do we need any tests/demo for this functionality?

wicketstuff-jquery-ui-parent/wicketstuff-jquery-ui/src/main/java/org/wicketstuff/jquery/ui/form/autocomplete/AutoCompleteTextField.java
```diff
@@ -313,7 +325,7 @@ protected CharSequence getChoiceCallbackUrl()
 				{
 					// warning, the template text should be of the form <a>...</a> in order to work
 					String render = "jQuery('%s').data('ui-autocomplete')._renderItem = function( ul, item ) { " // lf
-							+ "var content = jQuery.tmpl(jQuery('#%s').html(), item);" // lf
+							+ "var content = " + templateBehavior.getTemplateRenderingCode() // lf
```
martin-g: Shouldn't this be protected against NPEs ?
Both `templateBehavior` and `templateBehavior.getTemplateRenderingCode()` may return `null`.

reiern70: Checking

reiern70:
> LGTM, do we need any tests/demo for this functionality?

I will check. It seems to work Ok in our application in combination with custom values for autocomplete component.

wicketstuff-jquery-ui-parent/wicketstuff-jquery-ui/src/main/java/org/wicketstuff/jquery/ui/form/autocomplete/AutoCompleteTextField.java
reiern70: templateBehavior is not null.

```
protected String $()
			{
				if (templateBehavior != null)
				{
					// warning, the template text should be of the form <a>...</a> in order to work
					String render = "jQuery('%s').data('ui-autocomplete')._renderItem = function( ul, item ) { " // lf
							+ "var content = " + templateBehavior.getTemplateRenderingCode() // lf
							+ "return jQuery('<li/>').data('ui-autocomplete-item', item).append(content).appendTo(ul);" // lf
							+ "}";
					
					return super.$() + String.format(render, this.selector, templateBehavior.getToken());
				}

				return super.$();
			}
```

adding check for getTemplateRenderingCode

reiern70:
> LGTM, do we need any tests/demo for this functionality?

Added example

PR #1508, opened 2026-07-24
create a model based autocomplete component in order to allow avoidin…

reiern70:
…g serialising the choices.


reiern70:
I will merge this as we need it for our my customer project

solomax:
do we need these changes in `wicket-9.x`?

reiern70:
> do we need these changes in `wicket-9.x`?

I'm not sure we don't use 9.x for a long time. Our use case is we want to use an LDM for results and avoid serializing them with the page. No idea if this would be usefull to other users.

reiern70:
> do we need these changes in `wicket-9.x`?

I will create a PR. Not sue if there will be some conflict

Issue #711, opened 2021-09-13
wicketstuff-input-events - accumulating event listeners on AJAX components


reckart:
When adding an `InputBehavior` to a component which is updated via AJAX, it can happen that the event listener registered by the `InputBehavior` is registered over and over again during AJAX updates. This is a resource leak and also can lead to input events being triggered way too often.

I believe a reasonable fix for this behavior would be to update the JS which registers the listener to remove any existing listener for the key binding for which the new listener is created. This is may be as simple as adding a `shortcut.remove(keys)` before the shortcut.add(keys,...)` call in the template.

martin-g:
Pull Requests are welcome!

PR #1517, opened 2026-07-31
Issue #711: wicketstuff-input-events - accumulating event listeners on AJAX components

reckart:
- Fix keyboard shortcut listener accumulation on AJAX re-render by calling shortcut.remove() before shortcut.add() in three templates
- Switch InputBehavior.generateString() from in-place interpolate() to non-mutating asString(variables) to preserve component state across renders
- Add InputBehaviorRenderTest to verify scripts are re-generated from current component state per render
- Strengthen InputEventsTest with assertions on all three template types (explicit, autohook button, autohook link) to verify shortcut registration
- Delete vestigial WicketTestCase base class and inline its lifecycle into test subclasses
- Remove golden-file test fixtures (Page_expectedResult.html, test.html) that were never executed
- Enable test execution by removing surefire's "DoNotRunAnyTests" barrier and restoring inherited root configuration

(cherry picked from commit 867cc8fdd8e32e5f460b314816c315e905cc9eeb)


reckart:
@solomax Here's the Wicket 9.x cherry pick

